PR #23, opened 2022-01-27
feat: Slice support in CLI (WIP)

alexlokshin-czi:


cmd/create.go
```diff
@@ -106,7 +121,7 @@ func runCreate(cmd *cobra.Command, args []string) error {
 			return fmt.Errorf("failed to push image: %s", err)
 		}
 	}
-	stackMeta.Update(createTag, stackService)
+	stackMeta.Update(createTag, stackTags, sliceName, stackService)
```
edulop91, line 124: this probably should have an error check

cmd/create.go
```diff
@@ -85,6 +90,16 @@ func runCreate(cmd *cobra.Command, args []string) error {
...
+	var stackTags map[string]string
+	if len(sliceName) > 0 {
+		stackTags, createTag, err = buildSlice(happyConfig, sliceName, sliceDefaultTag)
+		if err != nil {
+			return err
+		}
+	} else {
+		stackTags = make(map[string]string)
+	}
+
```
edulop91, line 102:
```suggestion
	stackTags := make(map[string]string)
	if len(sliceName) > 0 {
		stackTags, createTag, err = buildSlice(happyConfig, sliceName, sliceDefaultTag)
		if err != nil {
			return err
		}
	}

```

edulop91: one less conditional to worry about

cmd/create.go
```diff
@@ -136,3 +151,48 @@ func runCreate(cmd *cobra.Command, args []string) error {
...
+	buildImages := slice.BuildImages
+	sliceTag, err := util.GenerateTag(happyConfig)
+	if err != nil {
+		return stackTags, defaultTag, err
+	}
+
+	err = runPushWithOptions(sliceTag, buildImages, "", "")
+	if err != nil {
+		return stackTags, defaultTag, fmt.Errorf("failed to push image: %s", err)
```
edulop91, line 176:
```suggestion
		return stackTags, defaultTag, fmt.Errorf("failed to push image: %w", err)
```
to wrap errors

cmd/create.go
```diff
@@ -136,3 +151,48 @@ func runCreate(cmd *cobra.Command, args []string) error {
...
 	return nil
 }
+
+func buildSlice(happyConfig config.HappyConfigIface, sliceName string, defaultSliceTag string) (stackTags map[string]string, defaultTag string, err error) {
+	stackTags = make(map[string]string)
+	defaultTag = defaultSliceTag
+
+	slices, err := happyConfig.GetSlices()
+	if err != nil {
+		return stackTags, defaultTag, fmt.Errorf("unable to retrieve slice configuration: %s", err)
```
edulop91, line 159:
```suggestion
		return stackTags, defaultTag, fmt.Errorf("unable to retrieve slice configuration: %w", err)
```
to wrap the error

cmd/list.go
```diff
@@ -72,11 +74,22 @@ var listCmd = &cobra.Command{
...
+			meta, err := stack.Meta()
 			if err != nil {
 				return err
 			}
-			tablePrinter.AddRow([]string{name, meta.DataMap["owner"], meta.DataMap["imagetag"], status, url})
+			tag := meta.DataMap["imagetag"]
+			imageTags, ok := meta.DataMap["imagetags"]
+			if ok && len(imageTags) > 0 {
+				var imageTagMap map[string]interface{}
+				json.Unmarshal([]byte(imageTags), &imageTagMap)
```
edulop91, line 85: does this need error check?

cmd/create.go
alexlokshin-czi: Agreed

cmd/list.go
alexlokshin-czi: Added a check

cmd/create.go
alexlokshin-czi: Added

PR #24, opened 2022-01-27
chore: Write silly tests to make coverage kick in

edulop91:
- write silly tests so coverage sits at > 0
- golangci lint so we make sure we do error check everywhere
- move argument checking conditionals to cli package
- do some error checks that were missing
- lint + autofmt


pkg/artifact_builder/artifact_registry.go
```diff
@@ -0,0 +1,12 @@
+package artifact_builder
```
edulop91, line 1: rename to snake_case file naming convention

alexlokshin-czi (approved):

PR #103, opened 2022-02-18
chore: Builder pattern and launch conditions validation

alexlokshin-czi:
…der()


pkg/util/preconditions.go
```diff
@@ -0,0 +1,49 @@
...
+	return errors.Wrap(err, "cannot connect to docker engine")
+}
+
+func RegistryLogin(userName string, password string, endpoint string) error {
+	client, err := client.NewClientWithOpts(client.FromEnv, client.WithAPIVersionNegotiation())
+	if err != nil {
+		return errors.Wrap(err, "docker engine is not running")
+	}
+
+	_, err = client.RegistryLogin(context.Background(), types.AuthConfig{
```
edulop91, line 43: registryLogin should take a ctx instead of using the context.background

pkg/util/preconditions.go
```diff
@@ -0,0 +1,49 @@
...
+	if err != nil {
+		return errors.Wrap(err, "could not find terraform in path")
+	}
+
+	client, err := client.NewClientWithOpts(client.FromEnv, client.WithAPIVersionNegotiation())
+	if err != nil {
+		return errors.Wrap(err, "docker engine is not running")
+	}
+
+	_, err = client.ContainerList(context.Background(), types.ContainerListOptions{})
```
edulop91, line 33: same here, pass in a ctx

pkg/util/preconditions.go
```diff
@@ -0,0 +1,49 @@
...
+	return errors.Wrap(err, "cannot connect to docker engine")
```
edulop91, line 34: I wonder if we should "collect" all these errors and return them as a combined error -- see `multierror` in other parts of the repo. That way it is easy for users to see all the things they should fix.

also, how about adding the ssm session manager check as well?

cmd/root.go
```diff
@@ -32,7 +33,10 @@ var rootCmd = &cobra.Command{
 			log.SetLevel(log.DebugLevel)
 			log.SetReportCaller(true)
 		}
-		return nil
+
+		err = util.ValidateEnvironment()
+
```
edulop91, line 38:
```suggestion
```

edulop91 (approved):

PR #106, opened 2022-02-23
fix: more consistently use slices; imagetags set properly

edulop91:


cmd/update.go
```diff
@@ -52,10 +52,19 @@ func runUpdate(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
-
 	builderConfig := artifact_builder.NewBuilderConfig().WithBootstrap(bootstrapConfig).WithHappyConfig(happyConfig)
-	ab := artifact_builder.NewArtifactBuilder().WithConfig(builderConfig).WithBackend(b)
+	buildOpts := []artifact_builder.ArtifactBuilderBuildOption{}
```
edulop91, line 56:
```suggestion
	buildOpts := []artifact_builder.ArtifactBuilderBuildOption{}
	// FIXME: this is an error-prone interface
```

pkg/artifact_builder/artifact_builder_options.go
```diff
@@ -20,6 +23,7 @@ func BuildSlice(slice *config.Slice) ArtifactBuilderBuildOption {
 // WithTags sets the tags that will be added to your docker image
 // Note that we will also set some default tags in addition
 func WithTags(tags ...string) ArtifactBuilderBuildOption {
+	log.Error(tags)
```
edulop91, line 26:
```suggestion
```

pkg/artifact_builder/builder_config.go
```diff
@@ -130,10 +130,14 @@ func (s *BuilderConfig) GetBuildServicesImage() (map[string]string, error) {
 
 	svcs := map[string]string{}
 	for serviceName, service := range configData.Services {
-		if service.Build != nil {
-			svcs[serviceName] = service.Image
+		// TODO: what is this service.Build?
```
edulop91, line 133:
```suggestion
		// NOTE: we assume for now docker-compose services without a build section are for local development only
```

pkg/stack_mgr/stack.go
```diff
@@ -104,7 +104,7 @@ func (s *Stack) Meta() (*StackMeta, error) {
 			return nil, err
 		}
 
-		// TODO(el): what is this?
+		// FIXME TODO(el): what is this?
```
edulop91, line 107:
```suggestion
		// FIXME TODO(el): why is this? don't we want to set these?
```

alexlokshin-czi (approved):

PR #118, opened 2022-02-24
fix!: match existing config key to auto_run_migrations

edulop91:
the python version uses `auto_run_migrations`, let's match it

### DIFF
--- a/pkg/config/happy_config.go
+++ b/pkg/config/happy_config.go
@@ -158,7 +158,7 @@ func (s *HappyConfig) GetSecretArn() string {
 	return envConfig.SecretARN
 }
 
-func (s *HappyConfig) AutoRunMigration() bool {
+func (s *HappyConfig) AutoRunMigrations() bool {
 	envConfig := s.getEnvConfig()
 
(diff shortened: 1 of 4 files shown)


alexlokshin-czi (approved):

pkg/config/happy_config.go
```diff
@@ -17,7 +17,7 @@ type Environment struct {
 	SecretARN          string     `yaml:"secret_arn"`
 	TerraformDirectory string     `yaml:"terraform_directory"`
 	DeleteProtected    bool       `yaml:"delete_protected"`
-	AutoRunMigration   bool       `yaml:"auto_run_migration"`
+	AutoRunMigrations  bool       `yaml:"auto_run_migrations"`
```
jgadling, line 20: Do we have a test that reads a config file and proves that all the values are set correctly?

edulop91: there should be, but probably missing this. updating

edulop91: added to the test

PR #143, opened 2022-03-04
feat: Added support for streaming of logs, 'get stack', and ability to force update when stack doesn't exist, fixes #137, #139

### LINKED ISSUE #137: feat: stream Terraform plan/apply output
When we kick of TFE ops, we should stream terraform output back to the users

### LINKED ISSUE #139: feat: command that checks if a stack exists or not
we do things like `(happy --profile="" list | grep -q $STACK_NAME` in a couple of places to determine if we should create or update a happy stack.

2 options:
- something like `happy get stack <stack_name>` will output information about the stack if it exists or an error otherwise. 
- `happy update` allows passing in a flag to create if missing.

alexlokshin-czi:
To retrieve a stack by name, run `happy get STACKNAME`, to create on update, if the stack doesn't exist, run `happy update STACKNAME --force`, it will update the stack if exists, or create if it doesn't. Logs are streamed when a workspace enters a "planning" or "applying" stage.


cmd/create.go
```diff
@@ -127,38 +130,69 @@ func runCreate(cmd *cobra.Command, args []string) error {
...
+		errs = multierror.Append(errs, errors.New("backend option not provided"))
 	}
+	if options.StackMeta == nil {
+		errs = multierror.Append(errs, errors.New("stackMeta option not provided"))
+	}
+	if len(options.StackName) == 0 {
+		errs = multierror.Append(errs, errors.New("stackName option not provided"))
+	}
+
+	if errs != nil {
```
edulop91, line 158: think this might be a problem with go's interface/nil compare

cmd/create.go
```diff
@@ -127,45 +130,79 @@ func runCreate(cmd *cobra.Command, args []string) error {
...
 
+	options := stackservice.NewStackManagementOptions(stackName).WithHappyConfig(happyConfig).WithStackService(stackService).WithStackMeta(stackMeta).WithBackend(backend)
+
 	// now that we have images, create all TFE related resources
-	secretArn := happyConfig.GetSecretArn()
-	if err != nil {
-		return err
+	err = createStack(ctx, options)
+
+	ab.Profiler.PrintRuntimes()
```
edulop91, line 138: we should `defer` this when we create the ab

pkg/workspace_repo/workspace.go
```diff
@@ -242,6 +267,35 @@ func (s *TFEWorkspace) WaitWithOptions(waitOptions options.WaitOptions) error {
 	return nil
 }
 
+func (s *TFEWorkspace) streamLogs(ctx context.Context, logs io.Reader) {
+	logrus.Info("...streaming logs...")
+	reader := bufio.NewReaderSize(logs, 64*1024)
```
edulop91, line 272: how about `bufio.Scanner`? I think it might give you a better level of abstraction

cmd/create.go
```diff
@@ -127,45 +131,78 @@ func runCreate(cmd *cobra.Command, args []string) error {
 		stackMeta = stackService.NewStackMeta(stackName)
 	}
 
+	options := stackservice.NewStackManagementOptions(stackName).WithHappyConfig(happyConfig).WithStackService(stackService).WithStackMeta(stackMeta).WithBackend(backend)
+
 	// now that we have images, create all TFE related resources
-	secretArn := happyConfig.GetSecretArn()
+	err = createStack(ctx, cmd, options)
+
+	return err
```
edulop91, line 139:
```suggestion
	return createStack(ctx, cmd, options)
```

cmd/get.go
```diff
@@ -0,0 +1,75 @@
...
+		}
+
+		stack, ok := stacks[stackName]
+		if !ok {
+			return errors.Errorf("stack '%s' not found in environment '%s'", stackName, happyConfig.GetEnv())
+		}
+
+		logrus.Infof("Retrieving stack '%s' from environment '%s'", stackName, happyConfig.GetEnv())
+
+		// TODO look for existing package for printing table
```
edulop91, line 62:
```suggestion
```

cmd/get.go
```diff
@@ -0,0 +1,75 @@
...
+		headings := []string{"Name", "Owner", "Tags", "Status", "URLs"}
+		tablePrinter := util.NewTablePrinter(headings)
+
+		err = stack.Print(ctx, stackName, tablePrinter)
+
```
edulop91, line 67:
```suggestion
```

cmd/get.go
```diff
@@ -0,0 +1,75 @@
...
+		if err != nil {
+			logrus.Errorf("Error retrieving stack %s:  %s", stackName, err)
```
edulop91, line 69: return this error

cmd/update.go
```diff
@@ -109,47 +114,88 @@ func runUpdate(cmd *cobra.Command, args []string) error {
...
+
+	options := stackservice.NewStackManagementOptions(stackName).WithHappyConfig(happyConfig).WithStackService(stackService).WithBackend(backend).WithStackTags(stackTags)
+
 	stack, ok := stacks[stackName]
 	if !ok {
-		return errors.Errorf("stack %s not found", stackName)
+		// Stack does not exist
+		if !force {
+			return errors.Errorf("stack '%s' does not exist, use --force or 'happy create %s' to create it", stackName, stackName)
+		} else { // Force creation of the new stack
```
edulop91, line 129:
```suggestion
		} 
		// Force creation of the new stack
```
you don't need the if/else here

cmd/update.go
```diff
@@ -109,47 +114,88 @@ func runUpdate(cmd *cobra.Command, args []string) error {
...
+			logrus.Infof("stack '%s' doesn't exist, it will be created", stackName)
+			stackMeta := stackService.NewStackMeta(stackName)
+			options = options.WithStackMeta(stackMeta)
+			return createStack(ctx, cmd, options)
+		}
```
edulop91, line 134:
```suggestion
```

cmd/update.go
```diff
@@ -109,47 +114,88 @@ func runUpdate(cmd *cobra.Command, args []string) error {
...
+	}
+
+	logrus.Infof("updating stack '%s'", stackName)
+	options = options.WithStack(stack)
+
+	// reset the configsecret if it has changed
+	// if we have a default tag, use it
+	err = updateStack(ctx, options)
+
+	return err
```
edulop91, line 144:
```suggestion
	return updateStack(ctx, options)
```

pkg/stack_mgr/stack.go
```diff
@@ -226,3 +228,33 @@ func (s *Stack) PrintOutputs(ctx context.Context) {
...
+	if err != nil {
+		return err
+	}
+	tag := meta.DataMap["imagetag"]
+	imageTags, ok := meta.DataMap["imagetags"]
+	if ok && len(imageTags) > 0 {
+		var imageTagMap map[string]interface{}
+		err = json.Unmarshal([]byte(imageTags), &imageTagMap)
+		if err != nil {
+			return err
```
edulop91, line 250: errors.wrap this one since it crosses the boundary between external code to our code

edulop91 (approved):
just a couple of quick syntax fixes, lgtm otherwise

PR #162, opened 2022-03-10
fix: docker-compose version contraints from panicing

jacobrheath:
This PR fixes a panic in the ValidateEnvironment function. Before, the API was panic-ing if the version wasn't right. Now, it throws an error. Also, it looked like the program was calling `docker compose` instead of `docker-compose`. I think the docker CLI only introduced `docker compose` in V2.  This PR fixes #162


pkg/util/preconditions.go
```diff
@@ -19,18 +19,21 @@ func ValidateEnvironment(ctx context.Context) error {
 	}
 
 	var errs *multierror.Error
-	_, err = exec.LookPath("docker")
+	_, err = exec.LookPath("docker-compose")
```
edulop91, line 22: we're using `docker compose` rather than `docker-compose` -- it is the direction the project is moving seems like

pkg/util/preconditions.go
```diff
@@ -19,18 +19,21 @@ func ValidateEnvironment(ctx context.Context) error {
...
+	v, err := exec.CommandContext(ctx, "docker-compose", "version", "--short").Output()
 	if err != nil {
 		errs = multierror.Append(errs, errors.Wrap(err, "could not determine docker compose version"))
 	}
 
 	version := strings.TrimSpace(string(v))
-	dockerComposeVersion := semver.MustParse(version)
+	dockerComposeVersion, err := semver.NewVersion(version)
+	if err != nil {
+		return errors.Wrap(err, "error getting the docker-compose version")
```
edulop91, line 35:
```suggestion
		return errors.Wrap(err, "error getting the docker compose version")
```

edulop91:
> This PR fixes a panic in the ValidateEnvironment function. Before, the API was panic-ing if the version wasn't right. Now, it throws an error. Also, it looked like the program was calling `docker compose` instead of `docker-compose`. I think the docker CLI only introduced `docker compose` in V2. This PR fixes #162

we require docker compose V2 for this project!

pkg/util/preconditions.go
jacobrheath: ooo duh ok. I will remove the docker-compose part.

jacobrheath: This is removed, but I have kept the function from panicing.

jacobrheath:
@edulop91 this makes more sense. I have fixed up this PR and left the code that prevents the panic.

edulop91 (approved):

Issue #164, opened 2022-03-10
StackNames must be compatible with DNS charset


edulop91:
We should check, upfront, that the provided StackNames in {create, delete, migrate, etc} are all compatible with DNS charset and error out immediately if they aren't

dtsai-czi:
To confirm, this character set is letters, digits, and hypens?

PR #168, opened 2022-03-11
feat: ensure that stack name inputs to CLI are compatible with DNS character set

dtsai-czi:
For #164


dtsai-czi:
https://czi.atlassian.net/wiki/spaces/SI/pages/2050588714/Happy+CLI#Warnings
Just noticed this warning, gonna add the length check to the PR and update this entry with the other checks implemented

pkg/cmd/validate.go
```diff
@@ -18,3 +21,22 @@ func Validate(vs ...cobra.PositionalArgs) cobra.PositionalArgs {
 		return nil
 	}
 }
+
+func CheckStackName(cmd *cobra.Command, args []string) error {
+	if notOk, err := stackNameIsInDnsCharset(args[0]); err != nil || notOk {
```
edulop91, line 26: I would also add a check that len(args) > 0, otherwise this could panic

edulop91, line 26: I would add a check here to make sure args has at least one element, otherwise this will panic. The way the error catching logic works, we consume all checks and aggregate the errors so this might potentially trigger a panic if someone runs the command without any args

edulop91 (approved):

PR #191, opened 2022-03-24
chore: Upgrade tfe client to v1.0.0 and other dependencies

alexlokshin-czi:


edulop91 (approved):

go.mod
```diff
@@ -1,22 +1,22 @@
 module github.com/chanzuckerberg/happy
 
-go 1.17
+go 1.18
```
edulop91, line 3: I think you also need to change the version in the github action files

edulop91 (changes requested):

edulop91 (approved):

PR #199, opened 2022-03-29
feat: "happy get [STACK]" to return extended details to improve navigation, closes #197

### LINKED ISSUE #197: feat: "happy get [STACK]" to return extended details to improve navigation
Display non-sensitive data from the integration secret, if stack is provided, list out key AWS resource ARNs or names. If possible, display fully qualified urls to easily access resources in AWS console.

alexlokshin-czi:


cmd/get.go
```diff
@@ -64,10 +69,150 @@ var getCmd = &cobra.Command{
...
+				}
+			}
+		}
+
+		tablePrinter.Print()
+		return nil
 	},
 }
+
+func arn2ConsoleLink(b *backend.Backend, unparsedArn string) (string, error) {
```
edulop91, line 175: this should probably move down to a `pkg` and get its own tests

cmd/get.go
```diff
@@ -64,10 +69,150 @@ var getCmd = &cobra.Command{
...
+					for _, containerDefinition := range taskDefinition.ContainerDefinitions {
+						tablePrinter.AddRow("      Name", *containerDefinition.Name)
+						tablePrinter.AddRow("      Image", *containerDefinition.Image)
+
+						logStreamPrefix := *containerDefinition.LogConfiguration.Options["awslogs-stream-prefix"]
+						logGroup := *containerDefinition.LogConfiguration.Options["awslogs-group"]
+						logRegion := *containerDefinition.LogConfiguration.Options["awslogs-region"]
+						containerName := *containerDefinition.Name
+
+						link := fmt.Sprintf("https://%s.console.aws.amazon.com/cloudwatch/home?region=%s#logEventViewer:group=%s;stream=%s/%s/%s", logRegion, logRegion, logGroup, logStreamPrefix, containerName, taskId)
```
edulop91, line 163: is there a chance this is not properly URL encoded? should we use the `url` pkg to encode these?

cmd/get.go
```diff
@@ -41,10 +44,10 @@ var getCmd = &cobra.Command{
 			return err
 		}
 
-		url := b.Conf().GetTfeUrl()
-		org := b.Conf().GetTfeOrg()
+		tfeUrl := b.Conf().GetTfeUrl()
```
edulop91, line 47: 👍

pkg/backend/aws/ecs.go
```diff
@@ -14,6 +14,68 @@ import (
...
+	clusterARN := b.integrationSecret.ClusterArn
+	out, err := b.ecsclient.DescribeServicesWithContext(ctx, &ecs.DescribeServicesInput{
+		Cluster:  &clusterARN,
+		Services: []*string{serviceName},
+	})
+	if err != nil {
+		return nil, errors.Wrap(err, "cannot describe an ECS service")
+	}
+	if len(out.Services) == 0 {
+		return nil, errors.Wrap(err, "ECS service was not found")
```
edulop91, line 27:
```suggestion
		return nil, errors.New("ECS service was not found")
```
since error is nil here, wrapping it will actually also return nil

pkg/backend/aws/ecs.go
```diff
@@ -14,6 +14,68 @@ import (
...
+		Cluster: &clusterARN,
+		Tasks:   []*string{taskArn},
+	})
+
+	if err != nil {
+		return []*ecs.TaskDefinition{}, errors.Wrap(err, "cannot describe ECS tasks")
+	}
+	taskDefinitions := []*ecs.TaskDefinition{}
+	for _, task := range tasksResult.Tasks {
+		taskDefResult, _ := b.ecsclient.DescribeTaskDefinitionWithContext(
```
edulop91, line 58: shouldn't ignore errors

cmd/get.go
alexlokshin-czi: 💯

pkg/util/awsconsole.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	switch resourceArn.Service {
+	case "ecs":
+
+		awsConsoleUrl.Path = "/ecs/home"
+
+		resourceParts := strings.Split(resourceArn.Resource, "/")
+		if len(resourceParts) < 2 {
+			return "", errors.Wrapf(err, "ARN is not supported: %s", unparsedArn)
```
edulop91, line 50:
```suggestion
			return "", errors.Errorf("ARN is not supported: %s", unparsedArn)
```

edulop91: err is nil here, so wrapping will also be nil

pkg/util/awsconsole.go
```diff
@@ -0,0 +1,133 @@
...
+	case "secretsmanager":
+		resourceParts := strings.Split(resourceArn.Resource, ":")
+		resourceType := resourceParts[0]
+
+		awsConsoleUrl.Path = "/secretsmanager/home"
+
+		switch resourceType {
+		case "secret":
+			secretName := strings.ReplaceAll(url.QueryEscape(options.IntegrationSecretARN), "%", "%%")
+			return awsConsoleUrl.String() + "#" + fmt.Sprintf("!/secret?name=%s", secretName), nil
```
edulop91, line 90: why these replace and fmt.sprintfs? rather than directly adding the data to awsConsoleURL and having that generate?

pkg/config/integration_secret.go
```diff
@@ -24,6 +24,7 @@ type TfeSecret struct {
 }
 
 type IntegrationSecret struct {
+	SecretArn      string
```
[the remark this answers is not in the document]
alexlokshin-czi: I moved it to the backend itself, not sure if this is more elegant.

pkg/util/awsconsole.go
alexlokshin-czi: I tried that... It double encodes the question mark, and the link is broken. Instead of producing %2F, it produces %252F.

edulop91 (approved):

PR #200, opened 2022-03-29
fix: Don't refer to docker-compose anymore

jgadling:
We're using `docker compose` in the CLI rather than `docker-compose` so this updates error/help text and comments accordingly


edulop91 (approved):

edulop91:
hmm, not sure why these are not getting triggered for you

PR #208, opened 2022-03-30
fix: Update command suppports --create-tag flag like Create does

edulop91:
Currently, only create supports the `--create-tag` flag, Update should too.


alexlokshin-czi (approved):

atolopko-czi:
Thank you, @edulop91!

PR #216, opened 2022-04-01
fix: Get logs path from ECS instead of generating

### LINKED ISSUE #214: fix: Get logs path from ECS instead of generating

alexlokshin-czi:
Fixes #214, depends on the `alokshin/aws-sdk-go-v2` branch.


pkg/backend/aws/ecs.go
```diff
@@ -187,17 +195,111 @@ func (ab *Backend) getNetworkConfig() *types.NetworkConfiguration {
...
+				log.Info(*event.Message)
+			}
+			return nil
+		},
+	)
+	if err != nil {
+		return errors.Wrapf(err, "error streaming logs")
+	}
+
+	return nil
```
edulop91, line 300:
```suggestion
		return errors.Wrapf(err, "error streaming logs")
```

pkg/backend/aws/ecs.go
```diff
@@ -225,41 +327,67 @@ func (ab *Backend) getLogEventsForTask(
...
+	}
+
+	arnSegments := strings.Split(resourceArn.Resource, "/")
+	if len(arnSegments) < 3 {
+		return "", errors.Errorf("incomplete task ARN: '%s'", taskArn)
+	}
+	return arnSegments[len(arnSegments)-1], nil
+}
+
+func (ab *Backend) getLogGroupAndStream(taskDefinition ecstypes.TaskDefinition, taskId string, containerName string) (string, string, error) {
```
edulop91, line 366:
```suggestion
func (ab *Backend) getLogGroupAndStreamNames(taskDefinition ecstypes.TaskDefinition, taskId string, containerName string) (string, string, error) {
```
what do you think of renaming like that? Otherwise, I thought that we would stream the log messages in this function as well

cmd/get.go
```diff
@@ -128,49 +129,59 @@ var getCmd = &cobra.Command{
...
+				return errors.Errorf("error retrieving task definition for tasks '%v'", taskArns)
+			}
+			taskDefinitionMap := map[string]ecstypes.TaskDefinition{}
+			for _, taskDefinition := range taskDefinitions {
+				taskDefinitionMap[*taskDefinition.TaskDefinitionArn] = taskDefinition
+			}
+
+			tasks, err := b.GetTaskDetails(ctx, taskArns)
+			if err != nil {
+				return errors.Errorf("error retrieving task details for tasks '%s'", taskArns)
```
edulop91, line 143:
```suggestion
				return errors.Wrapf(err, "error retrieving task details for tasks '%s'", taskArns)
```

cmd/get.go
```diff
@@ -128,49 +129,59 @@ var getCmd = &cobra.Command{
...
+					tablePrinter.AddRow("      Image", *containerDefinition.Image)
+
+					logStreamPrefix := containerDefinition.LogConfiguration.Options[backend.AwsLogsStreamPrefix]
+					logGroup := containerDefinition.LogConfiguration.Options[backend.AwsLogsGroup]
+					logRegion := containerDefinition.LogConfiguration.Options[backend.AwsLogsRegion]
+					containerName := *containerDefinition.Name
+
+					consoleLink, err := util.Log2ConsoleLink(util.LinkOptions{Region: logRegion}, logGroup, logStreamPrefix, containerName, taskId)
+					if err != nil {
+						return errors.Errorf("unable to construct a cloudwatch link for container '%s'", containerName)
```
edulop91, line 179: wrap

pkg/backend/aws/ecs.go
```diff
@@ -187,17 +195,111 @@ func (ab *Backend) getNetworkConfig() *types.NetworkConfiguration {
...
+			return errors.Wrapf(err, "invalid task ARN: '%s'", taskArn)
+		}
+
+		task := taskMap[taskArn]
+		taskDefinition := taskDefinitionMap[*task.TaskDefinitionArn]
+		logGroup, logStreamName, err = ab.getLogGroupAndStream(taskDefinition, taskId, containerName)
+		if err != nil {
+			log.Debugf("task definition %s does not have a log group", *taskDefinition.TaskDefinitionArn)
+			continue
+		}
```
edulop91, line 259: is this always true for this function when it returns an error? 
also, we lose contextual information from the error here and it is hidden

pkg/backend/aws/ecs.go
```diff
@@ -187,17 +195,111 @@ func (ab *Backend) getNetworkConfig() *types.NetworkConfiguration {
...
+	params := cloudwatchlogs.GetLogEventsInput{
+		LogGroupName:  &logGroup,
+		LogStreamName: &logStreamName,
+		StartTime:     aws.Int64(startTime.UnixNano() / int64(time.Millisecond)),
+		EndTime:       aws.Int64(endTime.UnixNano() / int64(time.Millisecond)),
+	}
+
+	log.Infof("Tailing logs: group=%s, stream=%s", logGroup, logStreamName)
+
+	err = ab.GetLogs(
```
edulop91, line 283:
```suggestion
	return ab.GetLogs(
```

edulop91 (approved):
lgtm, just a couple of small changes

pkg/backend/aws/ecs.go
alexlokshin-czi: I've added some context. Original code using the aws cli would stream logs without specifying the stream name:

```go
logPrefix := s.backend.Conf().LogGroupPrefix()
	logPath := fmt.Sprintf("%s/%s/%s", logPrefix, stackName, service)

	awsProfile := s.backend.Conf().AwsProfile()
	regionName := "us-west-2"
	awsArgs := []string{"aws", "--profile", *awsProfile, "--region", regionName, "logs", "tail", "--since", since, "--follow", logPath}
```
This code grabs the first valid stream.

PR #224, opened 2022-04-07
feat: Build amd64 images explicitly to avoid cross-platform issues

### LINKED ISSUE #204: feat: Build amd64 images explicitly to avoid cross-platform issues
Supposedly passing `--platform linux/amd64` into `docker build` and `docker run` will do.

alexlokshin-czi:
Closes #204


pkg/artifact_builder/artifact_builder.go
```diff
@@ -192,7 +193,28 @@ func (ab *ArtifactBuilder) RetagImages(
...
+					// Passing DOCKER_DEFAULT_PLATFORM into the docker process conflicts with the platform specified in the docker-compose.ymml itself
+					return errors.Errorf("Service '%s' has a platform '%s' specified in docker compose file, it is not possible to build the image. Please remove this setting.", serviceName, service.Platform)
+				}
+				targetPlatformDefinedInDockerCompose = true
+			}
+		}
+	}
+
+	// If targetPlatformDefinedInDockerCompose is true,it means that target platform matches the one define in happy config,
+	// and we can skip setting the DOCKER_DEFAULT_PLATFORM env variable on build.
```
edulop91, line 216: I would move all of this into a separate function call to encapsulate platform checking

pkg/artifact_builder/docker_compose.go
```diff
@@ -47,7 +51,14 @@ func (bc *BuilderConfig) invokeDockerCompose(command DockerCommand) ([]byte, err
 
 	envVars := bc.GetBuildEnv()
 	envVars = append(envVars, os.Environ()...)
-	envVars = append(envVars, "DOCKER_BUILDKIT=1")
+
+	// Specifying platform in the docker compose conflicts with the DOCKER_DEFAULT_PLATFORM env var
```
edulop91, line 55: what does this conflict look like? an error? or is the env variable ignored in favor of that in the config?

edulop91, line 55: what does this conflict look like? an error? or is the env variable ignored in favor of that in the config?

pkg/util/platform.go
```diff
@@ -0,0 +1,30 @@
...
+	"fmt"
+
+	"github.com/containerd/containerd/platforms"
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+func normalizeContainerPlatfrom(platform v1.Platform) v1.Platform {
+	platform.OS = "linux"
+	return platforms.Normalize(platform)
+}
```
edulop91, line 13: why setting this to linux?

edulop91, line 13: why setting this to linux?

pkg/util/platform.go
```diff
@@ -0,0 +1,30 @@
...
+
+func getUserContainerPlatform() v1.Platform {
+	platform := platforms.DefaultSpec()
+	return normalizeContainerPlatfrom(platform)
+}
+
+func GetSystemContainerPlatform(architecture string) string {
+	platform, err := platforms.Parse(architecture)
+	if err != nil {
+		return fmt.Sprintf("error: %s", err.Error())
```
edulop91, line 23: errors.Wrap

edulop91, line 23: errors.Wrap

pkg/artifact_builder/docker_compose.go
alexlokshin-czi: Config takes precedence, but env var is not ignored, `docker compose build` just exits with an error.

pkg/util/platform.go
alexlokshin-czi: All docker runtimes run images of `linux/amd64` or `linux/arm64` architecture -- even on `darwin` or `windows`.

alexlokshin-czi: Same as above. If you run `docker buildx ls`, you'll get a list of all supported runtimes by your machine; regardless of systems architecture, docker runtimes are all `linux`.

pkg/artifact_builder/artifact_builder_test.go
```diff
@@ -91,6 +91,9 @@ func TestCheckTagExists(t *testing.T) {
 	r.NoError(err)
 	err = artifactBuilder.Push([]string{"latest"})
 	r.NoError(err)
+
+	_, err = artifactBuilder.checkTargetPlatformDefinedInDockerCompose()
```
edulop91, line 95: why not also assert on the output here?

pkg/util/platform.go
edulop91: probably worth adding this in as a comment to the code

pkg/artifact_builder/docker_compose.go
```diff
@@ -47,7 +51,15 @@ func (bc *BuilderConfig) invokeDockerCompose(command DockerCommand) ([]byte, err
 
 	envVars := bc.GetBuildEnv()
 	envVars = append(envVars, os.Environ()...)
-	envVars = append(envVars, "DOCKER_BUILDKIT=1")
```
edulop91, line 50: why remove this one? seems like we actually want it and might resolve your error you mentioned in the other comment regarding buildkit?

edulop91:
going to close this one for now since we put the warning instead

PR #240, opened 2022-04-20
fix: create/update commands now use --use-existing-tag instead of --create-tag; more useful message when image verification fails

edulop91:


cmd/update.go
```diff
@@ -78,14 +84,14 @@ func runUpdate(cmd *cobra.Command, args []string) error {
 	stackService := stackservice.NewStackService().WithBackend(backend).WithWorkspaceRepo(workspaceRepo)
 
 	// build and push; creating tag if needed
-	if createTag && (tag == "") {
+	if !(skipTagCreation) && (tag == "") {
```
edulop91, line 87:
```suggestion
	if (!skipTagCreation) && (tag == "") {
```

alexlokshin-czi (approved):

cmd/create.go
```diff
@@ -48,15 +55,18 @@ func checkCreateFlags(cmd *cobra.Command, args []string) error {
 		return errors.New("--skip-check-tag can only be used when --tag is specified")
 	}
 
-	if !createTag && !cmd.Flags().Changed("tag") {
-		return errors.New("Must specify a tag when create-tag=false")
+	skipTagCreation := useExistingTag || (!createTag)
+	if skipTagCreation && !cmd.Flags().Changed("tag") {
+		return errors.New("Must specify a tag when use-existing-tag=true")
```
danieljhegeman, line 60: Oh, interesting. I was actually thinking of using it as `--use-existing-tag=<tag>` but semantically that would be incorrect. 

This solution is fine, defaults to false. 👍

danieljhegeman (approved):
thanks Eduardo

cmd/create.go
edulop91: that's also a good idea that we can use -- let me think a little bit about it. I'm not sure where the existing flags came to be and don't want to break some flow I'm not aware of

edulop91: I actually prefer your suggestion. Let me think a bit about it

Issue #268, opened 2022-05-02
bug: Happy CLI isn't running migrations by default after updating a stack


jgadling:
The CLI isn't running db migrations by default after updating a stack, and there's also no CLI option to disable this default action.

PR #269, opened 2022-05-03
fix: also run migrations on Update if requested

edulop91:
Fixes https://github.com/chanzuckerberg/happy/issues/268


jgadling:
Does this also include a `--skip-migrations/--do-migrations` to do the non-default behavior?
Screen Shot 2022-05-03 at 9 06 04 AM

edulop91:
> Does this also include a `--skip-migrations/--do-migrations` to do the non-default behavior? Screen Shot 2022-05-03 at 9 06 04 AM

no, but I can add that in another PR

jakeyheath (approved):

PR #270, opened 2022-05-03
feat: Warn if target compute platform is not specified in docker-compose.yml

alexlokshin-czi:
Fixes [CCIE-109]

### DIFF
--- a/cmd/hosts/install.go
+++ b/cmd/hosts/install.go
@@ -1,6 +1,8 @@
 package hosts
 
 import (
+	"context"
+
 	"github.com/chanzuckerberg/happy/pkg/artifact_builder"
 	"github.com/chanzuckerberg/happy/pkg/config"
 	"github.com/chanzuckerberg/happy/pkg/hostname_manager"
@@ -17,6 +19,7 @@ var installCmd = &cobra.Command{
 	Long:         "Install compose DNS entries",
 	SilenceUsage: true,
 	RunE: func(cmd *cobra.Command, args []string) error {
+		ctx := context.Background()
 		bootstrapConfig, err := config.NewBootstrapConfig(cmd)
 		if err != nil {
 			return err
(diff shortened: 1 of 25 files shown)


pkg/artifact_builder/builder_config.go
```diff
@@ -93,12 +98,26 @@ func (bc *BuilderConfig) retrieveConfigData() (*ConfigData, error) {
...
 
-func (s *BuilderConfig) GetConfigData() (*ConfigData, error) {
+func (bc *BuilderConfig) validateConfigData(ctx context.Context, configData *ConfigData) {
+	dctx, err := diagnostics.ToDiagnosticContext(ctx)
+	if err != nil {
+		log.Error("unable to create diagnostic context")
+	} else {
+		for serviceName, service := range configData.Services {
+			if len(service.Platform) == 0 {
+				dctx.AddWarning(fmt.Sprintf("service '%s' has no platform defined in docker-compose.yaml which can lead to unexpected side effects", serviceName))
```
edulop91, line 112: nice! I think I would flip this interface a bit and make it:
```
diagnostics.AddWarning(ctx, <the warning>)
```

cmd/root.go
```diff
@@ -24,6 +25,8 @@ func init() {
 	rootCmd.AddCommand(hosts.NewHostsCommand())
 }
 
+var ctx = diagnostics.BuildDiagnosticContext(context.Background())
```
edulop91, line 28:
```suggestion
// HACK: revert this singleton/global context once https://github.com/spf13/cobra/pull/1551 released
var ctx = diagnostics.BuildDiagnosticContext(context.Background())
```

cmd/hosts/install.go
```diff
@@ -27,7 +30,7 @@ var installCmd = &cobra.Command{
 		}
 
 		buildConfig := artifact_builder.NewBuilderConfig().WithBootstrap(bootstrapConfig).WithHappyConfig(happyConfig)
-		containers, err := buildConfig.GetContainers()
+		containers, err := buildConfig.GetContainers(ctx)
```
edulop91, line 33:
```suggestion
		containers, err := buildConfig.GetContainers(cmd.Context())
```
I think cmd has a ctx we should use but now don't remember

edulop91 (approved):

PR #272, opened 2022-05-03
feat: Allow overriding migration behavior via flags

edulop91:
fixes https://github.com/chanzuckerberg/happy/issues/268

introduces the following flags:
```
const (
	flagDoRunMigrations = "do-migrations"
	flagSkipMigrations  = "skip-migrations"
)
```
which can be specified to override the migration behavior defined in happy config. Note that we prevent both options from being specified at the same time and error out if so.

cc @jgadling


pkg/cmd/flags.go
```diff
@@ -20,3 +21,37 @@ func ValidateUpdateSliceFlags(cmd *cobra.Command, args []string) error {
...
+	}
+
+	if cmd.Flags().Changed(flagDoRunMigrations) {
+		run, err := cmd.Flags().GetBool(flagDoRunMigrations)
+		return run, errors.Wrapf(err, "could not read flag %s", flagDoRunMigrations)
+	}
+
+	if cmd.Flags().Changed(flagSkipMigrations) {
+		skip, err := cmd.Flags().GetBool(flagSkipMigrations)
+		run := !skip
```
jakeyheath, line 51: Why have this copy here?

pkg/cmd/flags.go
```diff
@@ -20,3 +21,37 @@ func ValidateUpdateSliceFlags(cmd *cobra.Command, args []string) error {
...
+
+		return run, errors.Wrapf(err, "could not read flag %s", flagSkipMigrations)
```
jakeyheath, line 53:
```suggestion
		return !skip, errors.Wrapf(err, "could not read flag %s", flagSkipMigrations)
```

jakeyheath, line 53:
```suggestion
		return !skip, errors.Wrapf(err, "could not read flag %s", flagSkipMigrations)
```

pkg/cmd/flags.go
```diff
@@ -20,3 +21,37 @@ func ValidateUpdateSliceFlags(cmd *cobra.Command, args []string) error {
...
+
+func SetMigrationFlags(cmd *cobra.Command) {
+	cmd.Flags().Bool("do-migrations", true, "Specify if you want to force migrations to run")
+	cmd.Flags().Bool("skip-migrations", false, "Specify if you want to skip migrations")
+}
+
+func ShouldRunMigrations(cmd *cobra.Command, happyConf *config.HappyConfig) (bool, error) {
+	if cmd.Flags().Changed(flagDoRunMigrations) && cmd.Flags().Changed(flagSkipMigrations) {
+		return false, errors.Errorf(
+			"Flags %s and %s cannot be specified at the same time",
```
jakeyheath, line 38: I think conventionally we shouldn't start errors with capital letters since they might be wrapped by other errors

jakeyheath (approved):
Just a couple nits

pkg/cmd/flags.go
edulop91: because sometimes these double negatives are confusing so I like to be explicit and give it a descriptive name for the reader. Doesn't have to be there, but I've found it helps avoid some types of bugs

pkg/cmd/flags.go
edulop91: I thought the linter caught these

edulop91, line 38:
```suggestion
			"flags %s and %s cannot be specified at the same time",
```

PR #289, opened 2022-05-12
fix: fix formatting of debug log for slow applies

jgadling:
- `log.Infof` doesn't need newlines added
- Only print the `many deregistered events` message once per crashing service, not once per log line
- Fix the happy script path in the suggested command to run.
- We want to show log lines *newer* than the last hour, not *older*


jgadling:
ref for previous output:
https://github.com/chanzuckerberg/czgenepi/runs/6409835519?check_suite_focus=true

alexlokshin-czi:
You need to either fix the coverage or `make coverage-update`

alexlokshin-czi (approved):

PR #296, opened 2022-05-17
chore: simplify logic for determining list of stack names

hspitzley-czi:


pkg/stack_mgr/stack_service.go
```diff
@@ -173,20 +169,17 @@ func (s *StackService) Add(ctx context.Context, stackName string) (*Stack, error
...
 	if err != nil {
 		return nil, err
 	}
 
-	for name := range existStacks {
-		existStackNames[name] = true
-	}
-	existStackNames[stackName] = true
 	newStackNames := []string{}
-	for name := range existStackNames {
```
hspitzley-czi, line 187: `existStackNames` was never used aside from the generation of `newStackNames`, and this simplifies that by removing the intermediate map

pkg/stack_mgr/stack_service.go
```diff
@@ -137,21 +137,17 @@ func (s *StackService) Remove(ctx context.Context, stackName string) error {
...
 	}
-	stackNames := map[string]bool{}
-	for currentStackName := range stacks {
-		if currentStackName != stackName {
-			stackNames[currentStackName] = true
-		}
-	}
 	stackNamesList := []string{}
-	for stackName := range stackNames {
-		stackNamesList = append(stackNamesList, stackName)
```
hspitzley-czi, line 148: `stackNames` was never used aside from the generation of `stackNamesList`, and this simplifies that by removing the intermediate map

pkg/stack_mgr/stack_service.go
```diff
@@ -163,7 +159,7 @@ func (s *StackService) Remove(ctx context.Context, stackName string) error {
 	if err != nil {
 		return errors.Wrap(err, "unable to resync the workspace")
 	}
-	delete(stacks, stackName)
```
hspitzley-czi, line 166: local `stacks` goes away after this but we probably want to keep the cache in sync

dtsai-czi (approved):

pkg/stack_mgr/stack_service.go
```diff
@@ -137,21 +137,17 @@ func (s *StackService) Remove(ctx context.Context, stackName string) error {
...
+		}
 	}
 
 	sort.Strings(stackNamesList)
 	stackNamesJson, err := json.Marshal(stackNamesList)
 	if err != nil {
-		return errors.Wrap(err, "unable to deserialize a stack list json")
+		return errors.Wrap(err, "unable to serialize stack list as json")
 	}
 	err = s.backend.WriteParam(ctx, s.writePath, string(stackNamesJson))
```
edulop91, line 152: not for this PR:
I think the order here is wrong. We should first delete from TFE and then delete from the stacklist. TFE delete should also probably wait. 

If we delete from stacklist and then TFE fails, we now have orphaned the stack.

edulop91 (approved):

PR #312, opened 2022-05-20
feat: List also prints the date last updated

edulop91:


edulop91:
it'll take some time for old dates to cycle out, but that might be fine

hspitzley-czi (approved):

PR #315, opened 2022-05-23
feat: use distributed dynamo locks to prevent stacklist race condition

hspitzley-czi:
This PR adds distributed locking which is necessary to prevent a race condition in [adding](https://github.com/chanzuckerberg/happy/blob/main/pkg/stack_mgr/stack_service.go#L171) and [removing](https://github.com/chanzuckerberg/happy/blob/main/pkg/stack_mgr/stack_service.go#L132) from the SSM stacklist in Happy which can lead to the `/happy/<env>/stacklist` SSM parameter getting out of sync with the `/happy/<env>/stacks/<name>` SSM parameters (leading to orphaned stacks). The locks are used to ensure nothing writes to the stacklist between when it is read and written to by a given process that acquired a lock.

This makes use of the DynamoDB table added in https://github.com/chanzuckerberg/shared-infra/pull/5318


pkg/stack_mgr/stack_service.go
```diff
@@ -164,6 +186,33 @@ func (s *StackService) Remove(ctx context.Context, stackName string) error {
...
+	_, err = distributedLock.ReleaseLock(ctx, lock)
+	if err != nil {
+		return nil, err
+	}
+
+	return stack, err
+}
+
+func (s *StackService) getDistributedLock() (*backend.DistributedLock, error) {
+	lockConfig := backend.DistributedLockConfig{DynamodbTableName: s.backend.Conf().DynamoLocktableName}
```
hspitzley-czi, line 212: Do we expect this lock to need a longer lease duration than the default of 10 seconds?

hspitzley-czi: Testing showed this was taking ~1.5 seconds for removing a stack and ~49s for adding a stack. I've refactored to reduce the time that a lock must be held for to keep it under a couple seconds

pkg/backend/aws/interfaces/mock_aws_dynamodb.go
```diff
@@ -0,0 +1,136 @@
+// Code generated by MockGen. DO NOT EDIT.
+// Source: pkg/backend/aws/interfaces/dynamodb.go
+
+// Package interfaces is a generated GoMock package.
```
hspitzley-czi, line 4: Generate by MockGen so this file doesn't need in-depth review

PR #328, opened 2022-06-06
fix: If the migration/deletion task succeeds too quickly, happy indicates failure status

alexlokshin-czi:
There are several different issues covered by this change, all caused by either task completing too quickly, or the log stream not being available in time. Instead of waiting on the task to be an indicator of a log being ready, waiting on the log stream itself. Fixes [CCIE-215].

[CCIE-215]: https://czi-eng.atlassian.net/browse/CCIE-215?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/backend/aws/ecs.go
```diff
@@ -216,12 +211,12 @@ func (ab *Backend) getNetworkConfig() *ecstypes.NetworkConfiguration {
 func (ab *Backend) Logs(ctx context.Context, stackName string, serviceName string, since string) error {
 	defer diagnostics.AddProfilerRuntime(ctx, time.Now(), "Logs")
 	endTime := time.Now()
+	var startTime *int64
 
 	duration, err := time.ParseDuration(since)
```
jakeyheath, line 216: We should probably handle the error here with a message of something like "that time format isn't supported"

jakeyheath (approved):
LGTM! Thanks for helping resolve that issue. Also, just one comment about an error.

PR #364, opened 2022-06-21
feat: Add `plan` support to the happy CLI

alexlokshin-czi:


cmd/create.go
```diff
@@ -205,42 +207,49 @@ func createStack(ctx context.Context, cmd *cobra.Command, options *stackservice.
...
 
-	shouldRunMigration, err := happyCmd.ShouldRunMigrations(cmd, options.HappyConfig)
-	if err != nil {
-		return err
-	}
-	if shouldRunMigration {
-		err = runMigrate(cmd, options.StackName)
+	if dryRun {
+		logrus.Infof("cleaning up stack '%s'", options.StackName)
+		err = options.StackService.Remove(ctx, options.StackName, true)
```
edulop91, line 226: why do we need to remove if this was a dry run?

alexlokshin-czi: Workspace has to be created even for a dry run, so it will be cleaned up after the dry run is complete. Workspace depends on TF variables that are not available locally, and even if they are their values can mismatch those set up by the creator workspace (i.e. `env-rdev`).

alexlokshin-czi: The only other way to handle it is to create a temp TFE workspace without the stack, and then clean it up afterwards; but that leads to a more complex execution path, as today TFE workspace is created by the creator workspace (i.e. `env-rdev`)

pkg/workspace_repo/workspace_repo.go
```diff
@@ -16,19 +16,20 @@ type Workspace interface {
 	WorkspaceName() string
 	GetCurrentRunID() string
 	GetLatestConfigVersionID() (string, error)
-	Run(isDestroy bool) error
+	Run(isDestroy bool, dryRun bool) error
```
jakeyheath, line 19: I know this would require a large refactor of what you have, but since we already have this interface, I wonder if instead of adding the dryRun bool to these methods, make a separate implementation of Workspace that is called something like `WorkspaceDryRunner` and when the dry-run flag is set in the command, `GetWorkspace()` returns the `WorkspaceDryRunner` that handles all the implementation details of running these functions in dry-run mode. Some psedocode:

~~~go
creatorWorkspace, err := s.workspaceRepo.GetWorkspace(ctx, s.creatorWorkspaceName, dryRun)

////


func (c *WorkspaceRepo) GetWorkspace(ctx context.Context, workspaceName string, dryRun bool) (Workspace, error) {
  if dryRun {
return WorkspaceDryRunner
}
return regularWorkspace
}
~~~

pkg/workspace_repo/workspace.go
```diff
@@ -406,11 +435,12 @@ func (s *TFEWorkspace) GetCurrentRunStatus() string {
 
 // create a new ConfigurationVersion in a TFE workspace, upload the targz file to
 // the new ConfigurationVersion, and finally return its ID.
-func (s *TFEWorkspace) UploadVersion(targzFilePath string) (string, error) {
+func (s *TFEWorkspace) UploadVersion(targzFilePath string, dryRun bool) (string, error) {
```
jakeyheath, line 438: See comment below about interface refactor.

pkg/workspace_repo/workspace.go
```diff
@@ -202,11 +215,11 @@ func (s *TFEWorkspace) RunConfigVersion(configVersionId string, isDestroy bool)
...
 }
 
-func (s *TFEWorkspace) Wait(ctx context.Context) error {
-	return s.WaitWithOptions(ctx, options.WaitOptions{})
+func (s *TFEWorkspace) Wait(ctx context.Context, dryRun bool) error {
+	return s.WaitWithOptions(ctx, options.WaitOptions{}, dryRun)
 }
 
-func (s *TFEWorkspace) WaitWithOptions(ctx context.Context, waitOptions options.WaitOptions) error {
+func (s *TFEWorkspace) WaitWithOptions(ctx context.Context, waitOptions options.WaitOptions, dryRun bool) error {
```
jakeyheath, line 222: See comment below about interface refactor.

pkg/workspace_repo/workspace.go
```diff
@@ -175,22 +184,26 @@ func (s *TFEWorkspace) SetVars(key string, value string, description string, sen
 	return errors.Wrapf(err, "could not create TFE variable %s:%s", key, value)
 }
 
-func (s *TFEWorkspace) RunConfigVersion(configVersionId string, isDestroy bool) error {
+func (s *TFEWorkspace) RunConfigVersion(configVersionId string, isDestroy bool, dryRun bool) error {
```
jakeyheath, line 187: See comment below about interface refactor.

pkg/workspace_repo/workspace.go
```diff
@@ -77,13 +86,13 @@ func (s *TFEWorkspace) GetLatestConfigVersionID() (string, error) {
 	return currentRun.ConfigurationVersion.ID, nil
 }
 
-func (s *TFEWorkspace) Run(isDestroy bool) error {
+func (s *TFEWorkspace) Run(isDestroy bool, dryRun bool) error {
```
jakeyheath, line 89: See comment below about interface refactor.

pkg/stack_mgr/stack_service.go
```diff
@@ -20,8 +20,8 @@ import (
 
 type StackServiceIface interface {
 	NewStackMeta(stackName string) *StackMeta
```
jakeyheath, line 22: I know this would require a large refactor of what you have, but since we already have this interface, I wonder if instead of adding the dryRun bool to these methods, make a separate implementation of `StackServiceIface` that is called something like `StackServiceDryRunner` and when the dry-run flag is set in the command, NewStackService() returns the `StackServiceDryRunner`, else it returns a normal `StackService`. `StackServiceDryRunner` would handle all the implementation details of running these functions in dry-run mode. Some psedocode:

~~~go
func NewStackService(dryRun bool) *StackService {
        if dryRun { 
              return &StackServiceDryRunner{}
        }
	// TODO pass this in instead?
	dirProcessor := util.NewLocalProcessor()

	return &StackService{
		stacks:       nil,
		dirProcessor: dirProcessor,
	}
}
~~~

pkg/stack_mgr/stack.go
```diff
@@ -125,41 +125,54 @@ func (s *Stack) Meta(ctx context.Context) (*StackMeta, error) {
...
 	workspace, err := s.getWorkspace(ctx)
 	if err != nil {
 		return err
 	}
-	return workspace.WaitWithOptions(ctx, waitOptions)
+	return workspace.WaitWithOptions(ctx, waitOptions, dryRun)
 }
 
-func (s *Stack) Apply(ctx context.Context, waitOptions options.WaitOptions) error {
+func (s *Stack) Apply(ctx context.Context, waitOptions options.WaitOptions, dryRun bool) error {
```
jakeyheath, line 167: See comment below about interface refactor.

pkg/stack_mgr/stack.go
```diff
@@ -125,41 +125,54 @@ func (s *Stack) Meta(ctx context.Context) (*StackMeta, error) {
 	return s.meta, nil
 }
 
-func (s *Stack) Destroy(ctx context.Context) error {
+func (s *Stack) Destroy(ctx context.Context, dryRun bool) error {
```
jakeyheath, line 128: See comment below about interface refactor.

pkg/orchestrator/orchestrator.go
```diff
@@ -198,7 +198,11 @@ func (s *Orchestrator) TaskExists(ctx context.Context, taskType backend.TaskType
 
 // Taking tasks defined in the config, look up their ID (e.g. ARN) in the given Stack
 // object, and run these tasks with TaskRunner
-func (s *Orchestrator) RunTasks(ctx context.Context, stack *stack_mgr.Stack, taskType backend.TaskType) error {
+func (s *Orchestrator) RunTasks(ctx context.Context, stack *stack_mgr.Stack, taskType backend.TaskType, dryRun bool) error {
+	if dryRun {
+		return nil
+	}
+
```
jakeyheath, line 205:
```suggestion
func (s *Orchestrator) RunTasks(ctx context.Context, stack *stack_mgr.Stack, taskType backend.TaskType) error {
	if s.dryRun {
		return nil
	}

```

I would make the dryRun bool a property of Orchestrator. Then add:

~~~go

func (s *Orchestrator) WithDryRun(dryRun bool) *Orchestrator {
	s.dryRun = dryRun
	return s
}
~~~

jakeyheath (changes requested):
Since we have interfaces in place for a lot of these structs, I don't think we should add the dryRun bool as a parameter to these functions. I think we should have a separate implementation of these interfaces that handles these methods when in dryRun mode. This is an easier to think about abstraction and takes the dryRun code/conditionals out of the regular business logic.

pkg/workspace_repo/workspace_repo.go
alexlokshin-czi: This is where the fundamental issue is. The difference between the dry run and a regular run is a `speculative` flag on the configuration, rest of the code in workspace.go is the same. I feel like creating a separate implementation will just duplicate the code, `speculative` is the only parameter that varies between the two.

pkg/orchestrator/orchestrator.go
alexlokshin-czi: Done.

pkg/workspace_repo/workspace_repo.go
jakeyheath: That's fair! Ignore my comments then

jakeyheath (approved):

PR #367, opened 2022-06-22
fix: fix image retagging (missing media formats, and registry id)

alexlokshin-czi:


pkg/artifact_builder/artifact_builder.go
```diff
@@ -150,15 +150,26 @@ func (ab *ArtifactBuilder) RetagImages(
 			}
 		}
 
-		repoUrl := strings.Split(registry.GetRepoUrl(), "/")[1]
+		parts := strings.Split(registry.GetRepoUrl(), "/")
```
edulop91, line 153: worth adding a test for this?

alexlokshin-czi: There's one already, but it didn't detect the issue.

PR #435, opened 2022-08-02
fix: (CCIE-452) Happy migrate in single-cell-data-portal errors out waiting for the log stream

alexlokshin-czi:


pkg/backend/aws/ecs.go
```diff
@@ -369,6 +360,33 @@ func (ab *Backend) getLogEventsForTask(
...
+				break
+			}
+		}
+	}()
+
+	var tasksResult *ecs.DescribeTasksOutput
+	select {
+	case res := <-ch:
+		tasksResult = res
+	case <-time.After(60 * time.Second):
```
jakeyheath, line 386: I think 1 * time.minute might be more readable.

pkg/backend/aws/ecs.go
```diff
@@ -369,6 +360,33 @@ func (ab *Backend) getLogEventsForTask(
 		return errors.Wrap(err, "could not describe task definition")
 	}
 
+	ch := make(chan *ecs.DescribeTasksOutput, 1)
+	go func() {
+		counter := 0
+		for {
```
jakeyheath, line 366:
```suggestion
		for now := range time.Tick(time.Minute) {
```

then you can remove your sleep on line 371

pkg/backend/aws/ecs.go
```diff
@@ -369,6 +360,33 @@ func (ab *Backend) getLogEventsForTask(
 		return errors.Wrap(err, "could not describe task definition")
 	}
 
+	ch := make(chan *ecs.DescribeTasksOutput, 1)
+	go func() {
```
jakeyheath, line 364: We need to kill this goroutine if the timeout hits

jakeyheath: Its possible we don't need this goroutine. You could move this logic to line 385 and wrap the select statement in a for loop

jakeyheath (approved):

PR #478, opened 2022-08-16
fix: log stream when there is no log prefix set

jakeyheath:
This PR fixes https://czi-tech.atlassian.net/browse/CCIE-503 where if the awslog-stream-prefix isn't set (like in [this task definition](https://github.com/chanzuckerberg/single-cell-explorer/blob/acd8932ffe09d980d57c7ab50b12b05816aa2709/.happy/terraform/modules/service/main.tf#L80)), there is no stream to pick. By default, happy will grab the latest log stream from the group. We should fix this using standarized TF modules for task definitions so that people don't forget to set a log prefix. 

See https://docs.aws.amazon.com/AmazonECS/latest/developerguide/using_awslogs.html for more details:

> If you don't specify a prefix with this option, then the log stream is named after the container ID that's assigned by the Docker daemon on the container instance. Because it's difficult to trace logs back to the container that sent them with just the Docker container ID (which is only available on the container instance), we recommend that you specify a prefix with this option.

### DIFF
--- a/pkg/backend/aws/ecs.go
+++ b/pkg/backend/aws/ecs.go
@@ -292,7 +292,7 @@ func (ab *Backend) extractLogInfoForServiceTask(ctx context.Context, taskArns []
 	containerName := ""
 	// Look for the first task definition that has a log group for a matching container name
 	for _, taskArn := range taskArns {
-		taskId, err := ab.getTaskId(taskArn)
+		taskId, err := ab.getTaskID(taskArn)
 		if err != nil {
 			return "", "", errors.Wrapf(err, "invalid task ARN: '%s'", taskArn)
 		}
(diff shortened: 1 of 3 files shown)


pkg/backend/aws/ecs.go
```diff
@@ -341,85 +341,155 @@ func (ab *Backend) extractLogInfoForArbitraryTask(ctx context.Context, stackName
...
-	if tasksResult == nil {
-		return errors.New("Unable to discover a task, impossible to stream the logs")
+	if err != nil {
+		return err
 	}
-
-	if tasksResult == nil || len(tasksResult.Tasks) == 0 || len(*tasksResult.Tasks[0].TaskArn) == 0 {
-		return errors.Errorf("no matching tasks for task definition %s", taskDefARN)
+	if tasks == nil {
+		return errors.New("unable to discover a task, impossible to stream the logs")
```
dtsai-czi, line 385: Combine this error with the one from the branch below?

dtsai-czi (approved):

PR #479, opened 2022-08-17
fix: refactor and combine log streams

jakeyheath:
This PR does:

* fixes https://czi-tech.atlassian.net/browse/CCIE-503 where if the awslog-stream-prefix isn't set (like in [this task definition](https://github.com/chanzuckerberg/single-cell-explorer/blob/acd8932ffe09d980d57c7ab50b12b05816aa2709/.happy/terraform/modules/service/main.tf#L80)), there is no stream to pick. Now by default, happy will grab all the log streams. We should fix this using standarized TF modules for task definitions so that people don't forget to set a log prefix. 

See https://docs.aws.amazon.com/AmazonECS/latest/developerguide/using_awslogs.html for more details:
> If you don't specify a prefix with this option, then the log stream is named after the container ID that's assigned by the Docker daemon on the container instance. Because it's difficult to trace logs back to the container that sent them with just the Docker container ID (which is only available on the container instance), we recommend that you specify a prefix with this option.

* refactors some of the logging code to look at the log configuration of the tasks instead of assuming a specific log configuration. I think this will make the log functionality more stable. 
* combines log streams from multiple containers so that we capture all logs, not just those from one single log stream. We now use FilterLogStream instead of GetLogStream so we can combine

Here's an example of what the new log streams look like:
Screen Shot 2022-08-18 at 10 24 58 AM


jakeyheath:
Still need to add tests, but ready to review @alexlokshin-czi @dtsai-czi @edulop91

alexlokshin-czi (approved):

alexlokshin-czi:
Looks good, but the go code needs to be reformatted.

go.mod
```diff
@@ -93,9 +94,13 @@ require (
...
+	golang.org/x/net v0.0.0-20220812174116-3211cb980234 // indirect
+	golang.org/x/sys v0.0.0-20220818161305-2296e01440c6 // indirect
 	golang.org/x/term v0.0.0-20220722155259-a9ba230a4035 // indirect
 	golang.org/x/text v0.3.7 // indirect
 	golang.org/x/time v0.0.0-20220722155302-e5dcc9cfc0b9 // indirect
 )
+
+// there is a breaking change in this library that fails when we update a stack
+// https://github.com/hashicorp/go-tfe/issues/507
+exclude github.com/hashicorp/go-slug v0.10.0
```
jakeyheath, line 106: excluding version updates until https://github.com/hashicorp/go-tfe/issues/507 is resolved

PR #576, opened 2022-09-21
feat: make stacklist path configurable

hspitzley-czi:
this allows the stacklist path to be configured, thus circumventing the issue where happy apps in the same aws account try to use the same stacklist param

https://github.com/chanzuckerberg/happy-infra/pull/95 updates the API stacklist param as well as the TF module to set a custom prefix


pkg/stack_mgr/stack_service.go
```diff
@@ -58,9 +58,13 @@ func NewStackService() *StackService {
...
 func (s *StackService) WithBackend(backend *backend.Backend) *StackService {
-	writePath := fmt.Sprintf("/happy/%s/stacklist", backend.Conf().GetEnv())
 	creatorWorkspaceName := fmt.Sprintf("env-%s", backend.Conf().GetEnv())
-	s.writePath = writePath
+
+	s.writePath = backend.Conf().GetSsmStacklistParamPath()
+	if s.writePath == "" {
+		// use the default value if no custom path is set
+		s.writePath = fmt.Sprintf("/happy/%s/stacklist", backend.Conf().GetEnv())
+	}
```
alexlokshin-czi, line 67: Are you solving a namespacing issue?

alexlokshin-czi (approved):

pkg/stack_mgr/stack_service.go
hspitzley-czi: Yep, exactly. This will allow us to namespace stacklists which will allow multiple happy apps to live in the same aws account without stacklists conflicting

PR #586, opened 2022-09-27
feat: use namespaced stacklist

hspitzley-czi:


hspitzley-czi:
> How is it going to work with existing environments? Is it backwards compatible?

@alexlokshin-czi it is backwards compatible 😄. When the namespaced param doesn't exist it will use the old param path as a fallback and it will do a dual write to ensure that the new param exists in the future.

jakeyheath (approved):

PR #587, opened 2022-09-27
fix: (CCIE-707) De-duplicate stacks on force create

dtsai-czi:


pkg/stack_mgr/stack_service.go
```diff
@@ -292,10 +292,16 @@ func (s *StackService) addToStacklist(ctx context.Context, stackName string) err
 	}
 
 	newStackNames := []string{}
+	stackNameExists := false
 	for name := range existStacks {
 		newStackNames = append(newStackNames, name)
+		if name == stackName {
+			stackNameExists = true
```
alexlokshin-czi, line 299:
```suggestion
break
```

dtsai-czi: If we break here, then names that come after the existing one won't make it into the new stack list, will they?

alexlokshin-czi: You are right.

alexlokshin-czi (approved):

PR #628, opened 2022-10-07
feat: Implement an abstraction around GetEvents to support ECS and K8S

alexlokshin-czi:


cli/pkg/backend/aws/k8s_compute_backend.go
```diff
@@ -281,3 +287,70 @@ func (k8s *K8SComputeBackend) Shell(ctx context.Context, stackName string, servi
...
+		})
+		if err != nil {
+			return errors.Wrapf(err, "unable to retrieve a list of pods for deployment %s", deploymentName)
+		}
+
+		warnings := make([]corev1.Event, 0)
+
+		fieldSelector := fields.SelectorFromSet(fields.Set{
+			"involvedObject.name": deploymentName,
+			"type":                "Warning",
```
hspitzley-czi, line 310: Should this use the const `Warning` variable?

cli/pkg/backend/aws/k8s_compute_backend.go
```diff
@@ -281,3 +287,70 @@ func (k8s *K8SComputeBackend) Shell(ctx context.Context, stackName string, servi
...
+			FieldSelector: fieldSelector.String(),
+			TypeMeta:      v1.TypeMeta{Kind: "Deployment"},
+		})
+
+		warnings = append(warnings, events.Items...)
+
+		for _, pod := range pods.Items {
+			fieldSelector := fields.SelectorFromSet(fields.Set{
+				"involvedObject.name": pod.Name,
+				"type":                "Warning",
```
hspitzley-czi, line 323: Should this use the const `Warning` variable?

cli/pkg/backend/aws/k8s_compute_backend.go
```diff
@@ -281,3 +287,70 @@ func (k8s *K8SComputeBackend) Shell(ctx context.Context, stackName string, servi
...
 	return t.Safe(func() error { return exec.Stream(streamOptions) })
 }
+
+func (k8s *K8SComputeBackend) GetEvents(ctx context.Context, stackName string, services []string) error {
+	if len(services) == 0 {
+		return nil
+	}
+
+	for _, serviceName := range services {
+		deploymentName := fmt.Sprintf("%s-%s", stackName, serviceName)
```
hspitzley-czi, line 297: Might be good to create an abstraction for deploymentName since it's used in a number of funcs in this file

hspitzley-czi: Maybe the whole instantiation of the `pods` variable should be abstracted since that's where deploymentName is used

cli/pkg/backend/aws/k8s_compute_backend.go
```diff
@@ -281,3 +287,70 @@ func (k8s *K8SComputeBackend) Shell(ctx context.Context, stackName string, servi
...
+		for _, e := range warnings {
+			logrus.Warnf("%s/%s - %s: %s", e.InvolvedObject.Kind, e.InvolvedObject.Name, e.Reason, e.Message)
+			warnings = append(warnings, e)
+		}
+
+		if len(events.Items) >= 1 {
+			logrus.Println()
+			logrus.Println("Many \"Warning\" events - please check to see whether your service is crashing:")
+			logrus.Infof("  happy --env %s logs %s %s", k8s.Backend.Conf().GetEnv(), stackName, serviceName)
+		}
```
hspitzley-czi, line 352: Probably not something to tackle now but separating the data retrieval from the data presentation would be good longer-term and would allow all backend computes to use the same presentation logic. I think we'll also need that separation to make it possible to move this logic to the API where it will return json and the CLI will need to parse and present it

cli/pkg/backend/aws/ecs_compute_backend.go
```diff
@@ -465,3 +466,69 @@ func (b *ECSComputeBackend) getTaskID(taskARN string) (string, error) {
...
+
+	clusterArn := b.Backend.Conf().GetClusterArn()
+
+	ecsClient := b.Backend.GetECSClient()
+
+	ecsServices := make([]string, 0)
+	for _, service := range services {
+		ecsService := fmt.Sprintf("%s-%s", stackName, service)
+		ecsServices = append(ecsServices, ecsService)
+	}
```
jakeyheath, line 483: I think we should avoid this convention throughout the codebase and move it into a function that is easier to change in the future. That is kind of what I did here: https://github.com/chanzuckerberg/happy/blob/main/cli/pkg/backend/aws/ecs.go#L20

alexlokshin-czi: This is a copy of previously existing code; I can wrap it in a function, but it is a bit out of scope for this pull request.

cli/pkg/backend/aws/k8s_compute_backend.go
```diff
@@ -184,14 +190,18 @@ func (k8s *K8SComputeBackend) WriteParam(
...
 func (k8s *K8SComputeBackend) PrintLogs(ctx context.Context, stackName string, serviceName string, opts ...util.PrintOption) error {
-	deploymentName := fmt.Sprintf("%s-%s", stackName, serviceName)
-	labelSelector := v1.LabelSelector{MatchLabels: map[string]string{"app": deploymentName}}
-	pods, err := k8s.ClientSet.CoreV1().Pods(k8s.HappyConfig.K8SConfig().Namespace).List(ctx, v1.ListOptions{
-		LabelSelector: labels.Set(labelSelector.MatchLabels).String(),
-	})
+	pods, err := k8s.getPods(ctx, stackName, serviceName)
+	if err != nil {
+		return errors.Wrap(err, "unable to retrieve a list of pods")
+	}
```
hspitzley-czi, line 201: error handling is duplicated
```suggestion
	pods, err := k8s.getPods(ctx, stackName, serviceName)
```

cli/pkg/backend/aws/k8s_compute_backend.go
```diff
@@ -281,3 +300,71 @@ func (k8s *K8SComputeBackend) Shell(ctx context.Context, stackName string, servi
...
+
+	for _, serviceName := range services {
+		pods, err := k8s.getPods(ctx, stackName, serviceName)
+		if err != nil {
+			return errors.Wrap(err, "unable to retrieve a list of pods")
+		}
+
+		if err != nil {
+			return errors.Wrap(err, "unable to retrieve a list of pods")
+		}
```
hspitzley-czi, line 317: error handling is duplicated
```suggestion
		pods, err := k8s.getPods(ctx, stackName, serviceName)
		if err != nil {
			return errors.Wrap(err, "unable to retrieve a list of pods")
		}
```

hspitzley-czi (approved):

PR #629, opened 2022-10-07
feat: auth in api

jakeyheath:
This PR adds authentication to the API. The auth needs to be configured by the app-config.yaml, but ideally we'd want to use env variables to configure such values. This PR adds the ability to use environment variables into the configuration file so that we can do both, but still have a single source of truth. 

Unrelatedly, this PR also fixes some race conditions in test files using `t.Parallel()` due to this issue: https://gist.github.com/posener/92a55c4cd441fc5e5e85f27bca008721


api/pkg/setup/setup.go
```diff
@@ -17,8 +17,10 @@ func init() {
 }
 
 type ApiConfiguration struct {
-	Port     uint   `mapstructure:"port"`
-	LogLevel string `mapstructure:"log_level"`
+	Port      uint   `mapstructure:"port"`
+	LogLevel  string `mapstructure:"log_level"`
+	IssuerURL string `mapstructure:"oidc_issuer_url"`
```
jakeyheath, line 22: open to ideas on how to improve this configuration. I imagine we don't want to hardcode the client IDs into a config file and commit it to source code. Can we use viper to pull environment variables @hspitzley-czi ?

api/main.go
```diff
@@ -20,7 +22,10 @@ func exec() error {
 	}
 	logrus.Info("Running with configuration:\n", string(m))
 
-	app := api.MakeApp(config)
+	app, err := api.MakeApp(context.Background(), config)
+	if err != nil {
+		return err
```
alexlokshin-czi, line 27: Please use `errors.Wrap()` to create a stack trace.

api/pkg/setup/setup.go
hspitzley-czi: You can do something like this: https://github.com/chanzuckerberg/happy/pull/634/files

Then set the values in `HAPPY_API_API_OIDC_CLIENT_ID` and `HAPPY_API_API_OIDC_ISSUER_URL` (note `API` appears twice, first because of the `happy_api` prefix and then again because the values are nested in the API config struct)

jakeyheath: so to deploy the API, we would need to have a configuration file and a set of environment variables? Or are you suggesting we use environment variables to fill in the values of the config file.

hspitzley-czi: If you merge the change in my PR we'll only need to set the values in `HAPPY_API_API_OIDC_CLIENT_ID` and  `HAPPY_API_API_OIDC_ISSUER_URL` env vars, viper will pick them up from there

jakeyheath: for reference, this is how the eng-portal config looks: https://github.com/chanzuckerberg/eng-portal/blob/0137f0f833c1caa33e9cfc3469ebf644bf9007c7/app-config.yaml#L74. The `${}` notation is used to add environment variables to the configuration file. What do you think of that?

hspitzley-czi: I tried briefly to do that with Viper and it didn't work, if you see a way to make that work I do think it would be preferred

api/Makefile
```diff
@@ -3,7 +3,7 @@ dev:
 	CONFIG_YAML_DIRECTORY=./ TZ=utc APP_ENV=development go run main.go
 
 test:
-	CONFIG_YAML_DIRECTORY=../.. TZ=utc APP_ENV=test go test -v ./... -run ^$(name)
+	CONFIG_YAML_DIRECTORY=../.. TZ=utc APP_ENV=test go test ./... -run ^$(name)
```
jakeyheath, line 6: removed this -v because it was really difficult to read which tests were failing

api/app-config.yaml
```diff
@@ -1,3 +1,6 @@
+auth:
+  oidc_client_id: "{{.OIDC_CLIENT_ID}}"
+  oidc_issuer_url: "{{.OIDC_ISSUER_URL}}"
```
jakeyheath, line 3: this is the notation we can use to drop in environment variables to our configuration files

api/main.go
```diff
@@ -1,26 +1,36 @@
...
 		return err
 	}
 	logrus.Info("Running with configuration:\n", string(m))
 
-	app := api.MakeApp(config)
+	cfg.Auth.Verifier, err = request.MakeOIDCVerifier(context.Background(), cfg.Auth.IssuerURL, cfg.Auth.ClientID)
+	if err != nil {
+		return err
+	}
+	app, err := api.MakeApp(cfg)
```
jakeyheath, line 30: I opted for this approach because all the other approaches were dangerously close to mixing production and test code together. This allows us to still use the configuration object, we just alter it a little bit before it goes into the makeapp call.

jakeyheath: There is also less "magic" string, which I think it hard to maintain. Its confusing unless we document it really well

api/pkg/api/app_test.go
```diff
@@ -50,13 +58,14 @@ func TestVersionCheckSucceed(t *testing.T) {
 	}
 
 	for idx, testCase := range testData {
+		tc := testCase
```
jakeyheath, line 61: fixed all these race condition things

api/pkg/setup/setup.go
```diff
@@ -35,12 +46,78 @@ type DatabaseConfiguration struct {
...
+		return nil, errors.Wrap(err, "unable to execute template")
+	}
+	return buff, nil
+}
+
+const defaultConfigYamlDir = "./"
+
 func GetConfiguration() (*Configuration, error) {
-	configYamlDir := os.Getenv("CONFIG_YAML_DIRECTORY")
+	configYamlDir := defaultConfigYamlDir
```
jakeyheath, line 117: maybe there is a more sane default?

api/pkg/request/auth.go
```diff
@@ -0,0 +1,80 @@
...
+	err = idToken.Claims(claims)
+	if err != nil {
+		return "", "", errors.Wrap(err, "unable to parse claims from ID token")
+	}
+	if len(claims.Email) == 0 {
+		return "", "", errors.New("missing email claim")
+	}
+	if len(claims.Subject) == 0 {
+		return "", "", errors.New("missing subject claim")
+	}
```
hspitzley-czi, line 59: What would you think of tags on the struct to indicate these as required? Similar to how the request payload validation works

jakeyheath: yea that might be nice. I don't know if they would be honored by this library though. Let me see

PR #657, opened 2022-10-12
feat: move models to shared package

hspitzley-czi:


api/go.mod
```diff
@@ -78,4 +79,7 @@ require (
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 )
 
-replace github.com/chanzuckerberg/happy/shared v0.0.0 => ../shared
+// this version was causing errors in the sqlite driver (eg: "cannot convert 0 (untyped int constant) to *int")
```
jakeyheath, line 82: if there is an open issue, we should reference in this comment

jakeyheath (approved):

api/go.mod
hspitzley-czi: I couldn't find an issue for this

PR #660, opened 2022-10-13
feat: roll out config feature to cli

hspitzley-czi:


shared/client/client.go
```diff
@@ -0,0 +1,100 @@
...
+
+func (c *HappyClient) Do(req *http.Request) (*http.Response, error) {
+	if c.clientVersion != "undefined" {
+		req.Header.Add("User-Agent", fmt.Sprintf("%s/%s", c.clientName, c.clientVersion))
+	}
+	req.Header.Add("Content-Type", "application/json")
+
+	// fmt.Println("headers:", req.Header)
+
+	// TODO: add auth header
```
hspitzley-czi, line 97: Will address this later

shared/client/util.go
```diff
@@ -0,0 +1,47 @@
...
+	defer resp.Body.Close()
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return errors.Wrap(err, "failed to read response body")
+	}
+
+	err = json.Unmarshal(body, &result)
+	if err != nil {
+		return errors.Wrap(err, "failed to unmarshal response body")
+	}
```
jakeyheath, line 25:
```suggestion
	err = json.NewDecoder(resp.Body).decode(&result)
	if err != nil {
		return errors.Wrap(err, "failed to unmarshal response body")
	}
```

Making a decoding is more efficent. It allows you to decode the JSON as you are reading the HTTP stream rather than reading all the bytes from the stream and then reading them all again to do the JSON decoding.

shared/client/util.go
```diff
@@ -0,0 +1,47 @@
...
+
+	return nil
+}
+
+func InspectForErrors(resp *http.Response) error {
+	if resp.StatusCode == http.StatusNotFound {
```
jakeyheath, line 31: switch case might be a bit more readable imo

shared/client/util.go
```diff
@@ -0,0 +1,47 @@
...
+		return ErrRecordNotFound
+	} else if resp.StatusCode == http.StatusBadRequest {
+		validationErrors := []model.ValidationError{}
```
jakeyheath, line 34: if you make the `model.ValidationError` implement the error interface, you could use the multierror package to format these errors in a little bit cleaner:

type error interface {
    Error() string
}

jakeyheath (approved):

shared/client/util.go
hspitzley-czi: Nice, good call 👍

shared/client/util.go
```diff
@@ -0,0 +1,47 @@
...
+		ParseResponse(resp, &validationErrors)
+		message := ""
+		for _, validationError := range validationErrors {
+			message = message + fmt.Sprintf("\nhappy-api request failed with: %s", validationError.Message)
+		}
+		return errors.New(message)
```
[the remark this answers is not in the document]
hspitzley-czi: added in [45342fd](https://github.com/chanzuckerberg/happy/pull/660/commits/45342fdea71ae13840fb8da4fb41c7b035c3db6d)

PR #661, opened 2022-10-14
chore: Cleanup naming patterns of modules and references inside of a mono repo to reflect reality of location

cyberious:
- Change pathing for module and references to code to meet golang standards
- Move modules referencing happy-api to happy/api instead
- Add gitignore to ensure we are not accidentally adding the binary files from make build


cyberious:
Bazel was added as an accident, cleaning up and will reopen.

PR #686, opened 2022-10-21
feat: Kubernetes migration and deletion task support

alexlokshin-czi:


cli/pkg/backend/aws/k8s_compute_backend.go
```diff
@@ -203,38 +206,128 @@ func (k8s *K8SComputeBackend) PrintLogs(ctx context.Context, stackName string, s
...
+	}
+	return nil
+}
 
-		for reader.Scan() {
-			logrus.Info(string(reader.Bytes()))
-		}
-		logrus.Infof("... done streaming ...")
+func (k8s *K8SComputeBackend) streamPodLogs(ctx context.Context, pod corev1.Pod, follow bool) error {
+	logrus.Infof("... streaming logs from pod %s ...", pod.Name)
```
jakeyheath, line 218: I feel like these should be more debug/verbose messages

cli/pkg/backend/aws/k8s_compute_backend.go
```diff
@@ -203,38 +206,128 @@ func (k8s *K8SComputeBackend) PrintLogs(ctx context.Context, stackName string, s
...
+	if err != nil {
+		return errors.Wrapf(err, "unable to retrieve logs from pod %s", pod.Name)
 	}
+	defer logs.Close()
+	reader := bufio.NewScanner(logs)
+
+	for reader.Scan() {
+		logrus.Info(string(reader.Bytes()))
+	}
+	logrus.Infof("... done streaming ...")
```
jakeyheath, line 232: I feel like these should be more debug/verbose messages

cli/pkg/backend/aws/k8s_compute_backend.go
```diff
@@ -203,38 +206,128 @@ func (k8s *K8SComputeBackend) PrintLogs(ctx context.Context, stackName string, s
...
 
 func (k8s *K8SComputeBackend) RunTask(ctx context.Context, taskDefArn string, launchType config.LaunchType) error {
-	// TODO: not implemented
-	return errors.New("not implemented")
+	// Get the cronjob and create a job out of it
+	if len(taskDefArn) >= 2 {
+		if taskDefArn[0] == '"' && taskDefArn[len(taskDefArn)-1] == '"' {
+			taskDefArn = taskDefArn[1 : len(taskDefArn)-1]
+		}
+	}
```
jakeyheath, line 242: I feel like this should be the callers responsibility

cli/pkg/backend/aws/k8s_compute_backend.go
```diff
@@ -203,38 +206,128 @@ func (k8s *K8SComputeBackend) PrintLogs(ctx context.Context, stackName string, s
...
+	if err != nil {
+		return errors.Wrap(err, "unable to create a k8s job")
+	}
+
+	logrus.Info("Waiting for all the pods to start")
+	// Make sure pods have started
+	counter := 0
+	var pods *corev1.PodList
+	for {
+		time.Sleep(5 * time.Second)
```
jakeyheath, line 260: I think this helper would be a little nicer here: https://github.com/chanzuckerberg/happy/blob/558e9e003eb8312f8e2ea35234b81a031b9d27aa/cli/pkg/util/time.go#L24. It also gives you a timeout period so you don't run an infinite loop by accident.

jakeyheath: It reruns until either the timeout is hit or you return a nil error

jakeyheath: then you can remove the counter

cli/pkg/backend/aws/k8s_compute_backend.go
```diff
@@ -203,38 +206,128 @@ func (k8s *K8SComputeBackend) PrintLogs(ctx context.Context, stackName string, s
...
+				break
+			}
+		}
+		if !allPodsReady {
+			for _, pod := range pods.Items {
+				if pod.Status.Phase == corev1.PodPending {
+					allPodsReady = false
+					break
+				}
+				logrus.Infof("%s: %s, %s", pod.Name, pod.Status.Phase, pod.Status.Message)
```
jakeyheath, line 289: I think this is a debug message, or maybe should be removed

cli/pkg/backend/aws/k8s_compute_backend.go
```diff
@@ -203,38 +206,128 @@ func (k8s *K8SComputeBackend) PrintLogs(ctx context.Context, stackName string, s
...
+				allPodsReady = false
+				break
+			}
+		}
+		if !allPodsReady {
+			for _, pod := range pods.Items {
+				if pod.Status.Phase == corev1.PodPending {
+					allPodsReady = false
+					break
+				}
```
jakeyheath, line 288:
```suggestion
```

Copy/pasta

jakeyheath: unless, did you mean to only print the pending items?

cli/pkg/backend/aws/k8s_compute_backend.go
jakeyheath, line 289:
```suggestion
				logrus.Infof("%s: %s, %s", pod.Name, pod.Status.Phase, pod.Status.Message)
```

jakeyheath, line 289: Debug message or remove

cli/pkg/backend/aws/k8s_compute_backend.go
```diff
@@ -203,38 +206,128 @@ func (k8s *K8SComputeBackend) PrintLogs(ctx context.Context, stackName string, s
...
+		break
+	}
+
+	logrus.Infof("Found %d pods", len(pods.Items))
+	for _, pod := range pods.Items {
+		err = k8s.streamPodLogs(ctx, pod, true)
+		if err != nil {
+			logrus.Error(err.Error())
+		}
+	}
```
jakeyheath, line 303: TODO for later, but we should probably interleave these instead of appending them to one another

cli/pkg/backend/aws/k8s_compute_backend.go
```diff
@@ -375,3 +468,35 @@ func (k8s *K8SComputeBackend) Describe(ctx context.Context, stackName string, se
 	}
 	return description, nil
 }
+
+func (k8s *K8SComputeBackend) createJobFromCronJob(cronJob *batchv1.CronJob, jobName string) *batchv1.Job {
+	annotations := make(map[string]string)
+	annotations["cronjob.kubernetes.io/instantiate"] = "manual"
```
jakeyheath, line 474:
```suggestion
	annotations:= map[string]string {
	  {"cronjob.kubernetes.io/instantiate": "manual"}
	}
```

jakeyheath (commented):
Left some cleanup items

cli/pkg/backend/aws/k8s_compute_backend.go
alexlokshin-czi: I'd rather keep the source of logs; if not, one will end up hunting the logs by hand, this gives it a source of reference.

cli/pkg/backend/aws/k8s_compute_backend.go
alexlokshin-czi: Logs are definitely not debug level messages, streaming markers are needed as markers, it makes no sense to give them a different level.

cli/pkg/backend/aws/k8s_compute_backend.go
```diff
@@ -203,38 +206,128 @@ func (k8s *K8SComputeBackend) PrintLogs(ctx context.Context, stackName string, s
...
+		Follow: follow,
+	}).Stream(ctx)
+	if err != nil {
+		return errors.Wrapf(err, "unable to retrieve logs from pod %s", pod.Name)
 	}
+	defer logs.Close()
+	reader := bufio.NewScanner(logs)
+
+	for reader.Scan() {
+		logrus.Info(string(reader.Bytes()))
```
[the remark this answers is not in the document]
alexlokshin-czi: Let's do it in a separate PR.

cli/pkg/backend/aws/k8s_compute_backend.go
alexlokshin-czi: How do you successfully interrupt this when you have successfully accomplished what you needed before timeout?

cli/pkg/backend/aws/k8s_compute_backend.go
alexlokshin-czi: It's a bug in our existing code.

alexlokshin-czi: quotes are passed into both ecs and k8s implementations

cli/pkg/backend/aws/k8s_compute_backend.go
jakeyheath: you return a nil error

jakeyheath (approved):

PR #700, opened 2022-10-24
feat: Extract common interface from ComputeLogPrinter and use it in PrintLogs() for k8s implementation

alexlokshin-czi:


cli/pkg/util/compute_log_printer.go
```diff
@@ -76,7 +79,7 @@ func MakeECSComputeLogPrinter(logGroupName string, logStreamNames []string, opts
 	return &lp
 }
 
-func (lp *ECSComputeLogPrinter) log(fleo *cloudwatchlogs.FilterLogEventsOutput, err error) error {
+func (lp *ComputeLogPrinter) log(fleo *cloudwatchlogs.FilterLogEventsOutput, err error) error {
```
jakeyheath, line 82: what if change this log function so that it just takes an array of bytes and reuse it between PrintCloudWatch and PrintReader. alternatively, we could make the cloudwatch stuff implement a reader interface and reuse PrintReader.

jakeyheath (approved):

PR #731, opened 2022-11-02
feat: happy dns module

jakeyheath:
This is our first CDK module for happy! It doesn't do a lot yet, it just is a mirror of the DNS module in most happy applications. It simply takes in the ALB that was created with happy-env-ecs, the zone they are working with, the stack and app name and makes a record for the stack that points to their ALB.


terraform/modules/__tests__/main-test.ts
```diff
@@ -0,0 +1,48 @@
...
+import { AwsProvider } from "@cdktf/provider-aws/lib/provider";
+describe("happy DNS", () => {
+  it("creates a AWS Route53 record", () => {
+    const construct = Testing.synthScope((scope) => {
+      new HappyDNS(scope, "my-dns", "albName", "albZoneId", "zone.czi.technology", "zoneId", "stackName", "appName");
+    })
+    expect(construct).toHaveResourceWithProperties(Route53Record, {
+      type: "A",
+      name: "stackName-appName.zone.czi.technology",
+      zone_id: "zoneId", // weird, it changes from camelCase to snake_case otherwise the test will fail
```
hspitzley-czi, line 16: I think this is because node/TS standard is to use camelcase and TF standard is snakecase

hspitzley-czi (approved):
This looks good but I think we'll eventually need to push the `main.ts` into a `src/` subdirectory to help organize as this grows

PR #734, opened 2022-11-04
feat: creating happy service for CDK modules

jakeyheath:
@hspitzley-czi  for help explaining.

I am taking a pause on this PR. As I started to experiment more with the Terraform CDK, it wasn't solving the goals of this PR without introducing a lot of refactoring. I was quickly going down the path of universal happy config, which I think we should do and use CDK for, but only once we have the appetite for it. Here are some of the learnings:

1. CDK is a code generator so it introduces the concept of "code gen" time and "terraform plan/apply" time. If we introduce variables that are only known at "terraform plan/apply" time, we can't utilize typing, typescript control flow constructs and other typescript features. We only get to use Typescript at "code gen" time. This means, we either need to be all in on CDK to take advantage of all the CDK features. Making a hybrid HCL module written in Typescript (my original idea) won't be very helpful.
2. CDK will be difficult to use with the integration secret. The integration secret is pulled via a Data block, which means its value won't be known until "terraform plan/apply", after the code gen has already run. This means, we can't type the integration secret or use typescript features based on the values that are in the integration secret.
3. CDK will be awesome for configuring happy services in a universal configuration via JSON, YAML, or even a network request without writing any terraform HCL or CDK code. [This PR shows ](terraform/modules/dns/main.ts)an example of how this might work.

Overall, I think we should get more folks involved in this process and make a more conscious decision about CDK for config. I think the config feature will be a good place to take our first steps in CDK rather than refactoring our existing modules in CDK.


alexlokshin-czi (approved):

abiju-czi:
Are we going with typescript as the CDK language or is this just a comparison?

jakeyheath:
@abiju-czi for now, we are using typescript. The golang version was not well supported unfortunately

terraform/modules/__tests__/__snapshots__/ecs-task-def.ts.snap
```diff
@@ -0,0 +1,43 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
jakeyheath, line 1: you don't need to review these files

terraform/modules/service/lb.ts
```diff
@@ -0,0 +1,116 @@
...
+    readonly lbName: string
+    readonly lbZoneID: string
+    readonly baseZone: DataAwsRoute53Zone
+    readonly listener: LbListener
+    readonly recordName: string
+
+    constructor(scope: Construct, id: string, config: HappyLoadBalancerProps) {
+        super(scope, id)
+
+        const pet = new Pet(scope, "petlb", { prefix: makeName(config.meta) })
```
jakeyheath, line 79: this was my idea to make random identifiers for resources. what do you think? we'd use tags to put all the human readable stuff

PR #785, opened 2022-11-22
feat: send auth header in api requests

hspitzley-czi:


shared/go.mod
```diff
@@ -3,35 +3,13 @@ module github.com/chanzuckerberg/happy/shared
 go 1.19
 
 require (
-	github.com/chanzuckerberg/go-misc v0.0.0-20221117151707-47f8234dc20c
```
jakeyheath: this was bumping the version since the CLI depends on this `shared` module

jakeyheath: figured it out by using `go mod graph | grep keyring` to see what other people in the package were using that go-misc or keyring library

cli/pkg/config/happy_config.go
```diff
@@ -275,8 +281,18 @@ func (s *HappyConfig) GetFeatures() *Features {
...
 
-func (s *HappyConfig) GetHappyApiBaseUrl() string {
-	return s.getData().HappyApiBaseUrl
+func (s *HappyConfig) GetHappyApiConfig() HappyApiConfig {
+	apiConfig := s.getData().Api
+	if apiConfig.BaseUrl == "" {
+		apiConfig.BaseUrl = "https://hapi.hapi.prod.si.czi.technology"
+	}
+	if apiConfig.OidcClientID == "" {
+		apiConfig.OidcClientID = "0oa7owjlihple45jJ5d7"
```
jakeyheath, line 290: instead of hardcoding, I wonder if we should just fail if these values aren't set to begin with and auth is required

terraform/provider/pkg/provider/provider.go
```diff
@@ -2,13 +2,91 @@ package provider
...
+		return nil, errors.Wrap(err, "error reading private key")
+	}
+	signingKey, err := jwt.ParseRSAPrivateKeyFromPEM(b)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to parse RSA private key")
+	}
+	return &TFTokenProvider{
+		signingKey: signingKey,
+		issuer:     issuer,
+		audience:   fmt.Sprintf("https://czi-prod.okta.com/oauth2/%s/v1/token", authzID),
```
jakeyheath, line 39: TODO: make this more configurable.

jakeyheath (approved):

PR #809, opened 2022-11-30
feat: add route53 module

hspitzley-czi:


terraform/modules/happy-route53/main.tf
```diff
@@ -0,0 +1,25 @@
...
+  name = local.hosted_zone_name
+  tags = var.tags
+}
+
+data "aws_route53_zone" "base" {
+  name         = local.base_zone_name
+  private_zone = false
+  providers = {
+    aws = aws.czi-si
+  }
```
hspitzley-czi, line 16: @jakeyheath is this where the provider alias should go or will it need to be in the caller of the module like in https://github.com/chanzuckerberg/shared-infra/blob/5119b15a3fc05ab5abf734f2164af9e5c69f8435/terraform/envs/happy-tfe/okta-apps/happy-api.tf#L7-L9?

jakeyheath: hmm I think you can set up an alias inside a module, but I don't know if I have tried before.

hspitzley-czi: let's try and if it breaks i'll make a PR to fix it 🎉

jakeyheath (approved):

PR #839, opened 2022-12-07
fix: remove old files

jakeyheath:


alexlokshin-czi (approved):

dtsai-czi (commented):
Will the refs to branch `heathj/fixbugs` break after this is merged and the branch is deleted?

PR #872, opened 2022-12-16
feat: implement the provider get token using private key

jakeyheath:
Forked from https://github.com/chanzuckerberg/happy/pull/785


alexlokshin-czi:
Looks like fmt.Print() needs to be replaced by a logrus invokation; that code is unchanged, how did the linter miss it last time?...

jakeyheath:
This is a stacked PR @alexlokshin-czi , its not based on main, its based on a different branch that I waiting to merge

terraform/provider/pkg/provider/provider.go
```diff
@@ -41,6 +112,6 @@ func Provider() *schema.Provider {
 
 func configureProvider(ctx context.Context, d *schema.ResourceData) (interface{}, diag.Diagnostics) {
 	apiBaseUrl := d.Get("api_base_url").(string)
-	api := client.NewHappyClient("happy-provider", version.ProviderVersion, apiBaseUrl, TfTokenProvider{})
+	api := client.NewHappyClient("happy-provider", version.ProviderVersion, apiBaseUrl, TFTokenProvider{})
```
hspitzley-czi, line 115: i believe this should use `MakeTFTokenProvider` now, right?

PR #887, opened 2022-12-16
feat: implement verifier for github and terraform provider

jakeyheath:
This PR implements the API logic that verifiers if an ID token/JWT is valid and has the proper claims before allowing them through to the core logic. These cases are special for the Github Actions flow and the Terraform provider which will be a machine-to-machine authentication.


alexlokshin-czi (approved):

api/app-config.development.yaml
```diff
@@ -1,5 +1,11 @@
 auth:
-  enable: false
+  enable: true
+  oidc_providers:
+    - issuer_url: https://czi-prod.okta.com/oauth2/aus7jg49oduHuUSs25d7
+      client_id: test-test-test-authz
+    - issuer_url: https://token.actions.githubusercontent.com
```
jakeyheath, line 6: current example of allowing any github actions through that have the owner of "chanzuckerberg"

api/pkg/request/auth.go
```diff
@@ -24,11 +26,80 @@ type OIDCVerifier interface {
...
+
+	if claims.Issuer != g.issuer {
+		return errors.Errorf("github id token didn't have the expected issuer, expected %s got %s", claims.Issuer, g.issuer)
+	}
+
+	return nil
+}
+
+func (o *OIDCProvider) Verify(ctx context.Context, rawIDToken string) (*oidc.IDToken, error) {
+	idToken, err := o.oidcVerifier.Verify(ctx, rawIDToken)
```
jakeyheath, line 85: tokens have to be valid (they are not expired and they are properly signed).

api/pkg/request/auth.go
```diff
@@ -24,11 +26,80 @@ type OIDCVerifier interface {
...
+	if err != nil {
+		return nil, errors.Wrap(err, "couldn't verify the ID token")
+	}
+
+	err = o.claimsVerifier.MatchClaims(ctx, idToken)
```
jakeyheath, line 90: they have to also pass the configured claims check. If they don't have the right claims it won't work either. This is nice for Github Action flow where the claims contain which owner and repo the Github Action is coming from.

PR #894, opened 2022-12-20
feat: service account for terraform provider credentials

jakeyheath:


terraform/modules/happy-tfe-okta-service-account/main.tf
```diff
@@ -0,0 +1,50 @@
...
+  label = "${var.service_name}-${var.app_name}-service-account"
+}
+
+module "happy_app" {
+  source = "git@github.com:chanzuckerberg/shared-infra//terraform/modules/okta-app-oauth?ref=heathj/jwks"
+
+  okta = {
+    label         = local.label
+    redirect_uris = concat(["https://oauth.${var.app_name}.si.czi.technology/oauth2/callback"], var.redirect_uris)
+    login_uri     = var.login_uri == "" ? "https://oauth.${var.app_name}.si.czi.technology" : var.login_uri
```
hspitzley-czi, line 24: do these need the env in the URI as well?

jakeyheath: I was thinking the service user would be shared between happy envs (prod,staging,rdev) so that we could share the key across all workspaces. But maybe we should make one per env instead so that we can differentiate the clients so that, for instance, rdev can't edit the config of prod or something

alexlokshin-czi (approved):

terraform/modules/happy-tfe-okta-service-account/main.tf
hspitzley-czi: Yeah I'm in favor of one per env

PR #919, opened 2022-12-24
feat: add service account to happy env

jakeyheath:


cyberious:
We should look carefully at this pattern. We could abstract, where we do not pin to just Okta but an authentication provider. Not a blocker but want to start thinking in that pattern of interfaces and moving away from being overly opinionated tool.

PR #924, opened 2023-01-03
feat: inject app configs into happy stack

hspitzley-czi:


terraform/modules/happy-stack-ecs/versions.tf
```diff
@@ -4,6 +4,10 @@ terraform {
       source  = "hashicorp/aws"
       version = ">= 4.45"
     }
+    hapi = {
+      source  = "github.com/chanzuckerberg/happy/terraform/provider"
```
jakeyheath, line 8: let me look into this and see what the source should be

jakeyheath:
stale, duplicated by #1108

PR #930, opened 2023-01-05
fix: make empty additional_env_vars map sensitive to fix dynamic for_each

hspitzley-czi:


hspitzley-czi:
no longer needed after https://github.com/chanzuckerberg/happy/pull/934

PR #962, opened 2023-01-10
feat: send aws creds in requests to api

hspitzley-czi:
This is needed to create an SSM client to support reading stacklists as part of CCIE-926


alexlokshin-czi (approved):

alexlokshin-czi (approved):

cli/pkg/backend/aws/backend.go
```diff
@@ -215,6 +215,10 @@ func NewAWSBackend(
 	return b, nil
 }
 
+func (b Backend) GetCredentials() (aws.Credentials, error) {
+	return b.awsConfig.Credentials.Retrieve(context.Background())
```
jakeyheath, line 219: we should probably pass a context in here and then use the CLI's context that comes from that CLI library

cli/pkg/hapi/api.go
```diff
@@ -22,15 +25,32 @@ func (t CliTokenProvider) GetToken() (string, error) {
 	return token.IDToken, nil
 }
 
+type CliAwsCredentialsProvider struct {
```
jakeyheath, line 28: typically in go programs, acronyms are always uppercase. This gets a little weird when types have two accronyms right next to each other

https://stackoverflow.com/questions/39830166/glued-acronyms-and-golang-naming-convention

cli/pkg/hapi/api.go
```diff
@@ -22,15 +25,32 @@ func (t CliTokenProvider) GetToken() (string, error) {
...
+	happyConfig *config.HappyConfig
+}
+
+func (c CliAwsCredentialsProvider) GetCredentials() (aws.Credentials, error) {
+	b, err := backend.NewAWSBackend(context.Background(), c.happyConfig)
```
jakeyheath, line 33: pass context in as part of the GetCredentials call

terraform/provider/pkg/provider/provider.go
```diff
@@ -31,13 +31,7 @@ type PrivateKeyTFTokenProvider struct {
...
 }
 
-func MakeKMSKeyTFProvider(ctx context.Context, provConfig *Config) (client.TokenProvider, error) {
-	cfg, err := config.LoadDefaultConfig(ctx)
-	if err != nil {
-		return nil, errors.Wrap(err, "unable to load SDK config")
-	}
-	stsClient := sts.NewFromConfig(cfg)
-	appCreds := stscreds.NewAssumeRoleProvider(stsClient, *provConfig.AssumeRoleARN)
+func MakeKMSKeyTFProvider(ctx context.Context, provConfig *Config, appCreds *stscreds.AssumeRoleProvider) (client.TokenProvider, error) {
```
jakeyheath, line 34: did this need to change?

terraform/provider/pkg/provider/provider.go
```diff
@@ -147,6 +141,22 @@ func requestAccessToken(scope, audience, signedToken string) (string, error) {
 	return accessTokenResp.AccessToken, nil
 }
 
+type TfAwsCredentialsProvider struct {
```
jakeyheath, line 144: see comment about golang and accroynms

terraform/provider/pkg/provider/provider.go
```diff
@@ -147,6 +141,22 @@ func requestAccessToken(scope, audience, signedToken string) (string, error) {
...
+	appCreds *stscreds.AssumeRoleProvider
+	ctx      context.Context
```
jakeyheath, line 146: I think its an antipattern to pass a context as part of a struct attribute. The typical convention is to always include the context as part of the first argument of the function, like you do in getAWsCredsProvider

https://groups.google.com/g/golang-nuts/c/xRbzq8yzKWI?pli=1

terraform/provider/pkg/provider/provider.go
```diff
@@ -147,6 +141,22 @@ func requestAccessToken(scope, audience, signedToken string) (string, error) {
...
+}
+
+func (t TfAwsCredentialsProvider) GetCredentials() (aws.Credentials, error) {
+	return t.appCreds.Retrieve(t.ctx)
+}
+
+func getAwsCredsProvider(ctx context.Context, appCreds *stscreds.AssumeRoleProvider) client.AWSCredentialsProvider {
```
jakeyheath, line 153: acroynms

terraform/provider/pkg/provider/provider.go
```diff
@@ -147,6 +141,22 @@ func requestAccessToken(scope, audience, signedToken string) (string, error) {
 	return accessTokenResp.AccessToken, nil
 }
 
+type TfAwsCredentialsProvider struct {
+	appCreds *stscreds.AssumeRoleProvider
+	ctx      context.Context
+}
+
+func (t TfAwsCredentialsProvider) GetCredentials() (aws.Credentials, error) {
+	return t.appCreds.Retrieve(t.ctx)
```
jakeyheath, line 150: pass context in from GetCredenitals instead of struct attribute

jakeyheath (changes requested):
A few changes to using context and accroynms

terraform/provider/pkg/provider/provider.go
hspitzley-czi: 👍 I'll update `GetCredentials` to take a context arg then it wont be necessary as a struct field

terraform/provider/pkg/provider/provider.go
hspitzley-czi: I changed it so the same `stscreds.AssumeRoleProvider` object could be used here and in the `TfAwsCredentialsProvider`. Is there a reason they should use different `stscreds.AssumeRoleProvider` objects?

shared/client/client.go
```diff
@@ -110,9 +119,17 @@ func (c *HappyClient) Do(req *http.Request) (*http.Response, error) {
 func (c *HappyClient) addAuth(req *http.Request) error {
 	token, err := c.tokenProvider.GetToken()
 	if err != nil {
-		return errors.Wrap(err, "failed to get token")
+		return errors.Wrap(err, "failed to get oidc token")
 	}
 	req.Header.Add("Authorization", fmt.Sprintf("Bearer %s", token))
 
+	creds, err := c.awsCredProvider.GetCredentials(context.Background())
```
jakeyheath, line 126: I think this is also a place you need to pass the context in instead of using context.Background

terraform/provider/pkg/provider/provider.go
jakeyheath: oo I see. I think Github was hiding a few other lines that were deleted and I didn't notice what was going on

shared/client/client.go
hspitzley-czi: I think I'll leave this how it is for now since the client doesn't have a context being passed to it currently. We can thread that through later if necessary

PR #1008, opened 2023-01-24
feat: add synthetics to ecs stacks module

jakeyheath:
This module needs a lot of love, but this is the minimum required to add synthetics to this version of the module. We should make a separate PR that brings this interface as close to the EKS version one so that the drift doesn't go too far. For example, ECS right now only supports 1 service and has not service configuration like the EKS module does.


hspitzley-czi (approved):

hspitzley-czi (changes requested):
Looks like this check is failing https://github.com/chanzuckerberg/happy/actions/runs/4001518575/jobs/6867842984

dtsai-czi (approved):

hspitzley-czi (approved):

PR #1033, opened 2023-01-30
feat: Tag stack level metrics for EKS and ECS

alexlokshin-czi:


hspitzley-czi (approved):

terraform/modules/happy-env-ecs/integration_secret.tf
```diff
@@ -55,6 +55,7 @@ locals {
     }
 
     dynamo_locktable_name = aws_dynamodb_table.locks.id
+    datadog_api_key       = var.datadog_api_key
```
jakeyheath, line 58: hmm

terraform/modules/happy-service-ecs/main.tf
```diff
@@ -24,6 +24,111 @@ resource "aws_ecs_service" "service" {
...
+          hostPort      = 8125
+          protocol      = "udp"
+      }]
+
+      "logConfiguration" = {
+        logDriver = "awslogs"
+
+        options = {
+          awslogs-stream-prefix = var.app_name,
+          awslogs-group         = aws_cloudwatch_log_group.cloud_watch_logs_group.id,
```
jakeyheath, line 127: should we have a different log group for datadog so the app logs don't get mixed in with the datadog logs?

terraform/modules/happy-env-ecs/integration_secret.tf
alexlokshin-czi: We can't reference czi-si secrets in this repo... Btw, I did create one.

terraform/modules/happy-service-ecs/main.tf
alexlokshin-czi: Good point.

terraform/modules/happy-service-ecs/main.tf
```diff
@@ -24,6 +24,111 @@ resource "aws_ecs_service" "service" {
 
 locals {
   task_definition = [
+    {
+      name      = "datadog-agent"
+      essential = true
+      image     = "public.ecr.aws/datadog/agent:latest"
```
jakeyheath, line 30: can we make this configurable with these as the defaults? one object input variable that has the image_host and version? I have had too many cases where I needed to change the image version of something and I had to re-release the module. It's way easier to have it be configurable with sane defaults.

terraform/modules/happy-service-ecs/main.tf
```diff
@@ -24,6 +24,111 @@ resource "aws_ecs_service" "service" {
...
+      cpu       = 512
+      memory    = 512
+
+      environment = concat(
+        [
+          {
+            name  = "DD_API_KEY"
+            value = var.datadog_api_key
```
jakeyheath, line 38: this is probably okay for now, but I wonder if instead of value, we try to use something like this:

~~~
{
  "containerDefinitions": [{
    "secrets": [{
      "name": "environment_variable_name",
      "valueFrom": "arn:aws:secretsmanager:region:aws_account_id:secret:appauthexample-AbCdEf:username1::"
    }]
  }]
}
~~~

jakeyheath: https://aws.amazon.com/premiumsupport/knowledge-center/ecs-data-security-container-task/

jakeyheath (approved):

terraform/modules/happy-service-ecs/main.tf
```diff
@@ -24,6 +24,111 @@ resource "aws_ecs_service" "service" {
 
 locals {
   task_definition = [
+    {
+      name      = "datadog-agent"
+      essential = true
+      image     = "${var.datadog_image.registry}:${var.datadog_image.tag}"
+      cpu       = 512
```
jakeyheath, line 31: probably should add cpu and memory to the datadog image object too

terraform/modules/happy-service-ecs/main.tf
alexlokshin-czi: I don't think we should do it at the service level. Feeding it into the integration secret this way, perhaps.

PR #1068, opened 2023-02-03
feat: CCIE-926 List of all happy stacks for an app env

hspitzley-czi:


api/pkg/api/config_group.go
```diff
@@ -46,9 +46,18 @@ func getPayload[T interface{}](c *fiber.Ctx) T {
...
+// TODO: rename to parseRequestBody as part of https://czi-tech.atlassian.net/browse/CCIE-1005
 func parsePayload[T interface{}](c *fiber.Ctx) error {
+	return parsePayloadImpl[T](c, c.BodyParser)
+}
+
+func parseQueryString[T interface{}](c *fiber.Ctx) error {
+	return parsePayloadImpl[T](c, c.QueryParser)
+}
+
+func parsePayloadImpl[T interface{}](c *fiber.Ctx, fn request.RequestParser) error {
```
jakeyheath, line 58: I'm not a super fan of the naming convention of "impl" suffixes. Doesn't really describe what its doing

api/pkg/api/stack_group.go
```diff
@@ -2,26 +2,31 @@ package api
...
 	"github.com/chanzuckerberg/happy/api/pkg/cmd"
+	"github.com/chanzuckerberg/happy/api/pkg/request"
 	"github.com/chanzuckerberg/happy/api/pkg/response"
 	"github.com/chanzuckerberg/happy/shared/model"
 	"github.com/gofiber/fiber/v2"
 )
 
 type StackHandler struct {
-	stack cmd.Stack
+	stack cmd.StackIface
```
jakeyheath, line 12: same with Iface as part of the name. The type already shows its an interface, so I don't see the benefit for add "Iface" to it. Is there a particular reason for this naming choice?

api/pkg/cmd/stack.go
```diff
@@ -1,65 +1,61 @@
...
 	"github.com/pkg/errors"
-	"gorm.io/gorm/clause"
+	"github.com/sirupsen/logrus"
 )
 
-type Stack interface {
-	CreateOrUpdateAppStack(model.AppStackPayload) (*model.AppStack, error)
-	GetAppStacks(model.AppStackPayload) ([]*model.AppStack, error)
-	DeleteAppStack(model.AppStackPayload) (*model.AppStack, error)
+type StackIface interface {
```
jakeyheath, line 13: https://go.dev/doc/effective_go#interface-names

api/pkg/api/stack_group.go
```diff
@@ -2,26 +2,31 @@ package api
...
 }
 
-func MakeStackHandler(s cmd.Stack) *StackHandler {
+func MakeStackHandler(s cmd.StackIface) *StackHandler {
 	return &StackHandler{
 		stack: s,
 	}
 }
 
+func MakeStackHandler2() *StackHandler {
```
jakeyheath, line 21: I think this might have been leftover. Can you remove it?

api/pkg/cmd/stack.go
jakeyheath: StackGetter, StackSetter, StackDeleter, then combine those to be like StackManager or something

api/pkg/cmd/stack.go
```diff
@@ -1,65 +1,61 @@
...
-}
-
-func (s *dbStack) GetAppStacks(payload model.AppStackPayload) ([]*model.AppStack, error) {
-	db := s.DB.GetDB()
-	stack := &model.AppStack{AppStackPayload: payload}
-	stacks := []*model.AppStack{}
-	res := db.Where(stack).Find(&stacks)
-	return stacks, errors.Wrapf(res.Error, "unable to get app stacks for %s", stack.AppMetadata)
+func (s Stack) GetAppStacks(ctx context.Context, payload model.AppStackPayload) ([]*model.AppStack, error) {
+	switch payload.TaskLaunchType {
```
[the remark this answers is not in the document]
jakeyheath: alternatively, in MakeStack, do the switch case and instead of having a struct with all three (db, eks, ecs), only return the stack interface that matches.

jakeyheath (changes requested):
I think some of the interface stuff can be simplified and typed so that there is less string validation during the program execution time. I find having types instead of string comparisons is safer and harder to make mistakes and helps documentation better.

api/pkg/api/config_group.go
jakeyheath: trying to find a go reference for this comment one moment

hspitzley-czi: Fair, this is temporary though, I have another ticket to come make some more changes to this area and wanted to keep it separate since this is already a large PR

api/pkg/api/stack_group.go
hspitzley-czi: I wanted `Stack` reserved for the struct below. I'll rename the StackIface to StackManager

api/pkg/api/stack_group.go
hspitzley-czi: Good catch, removed

api/pkg/cmd/stack.go
hspitzley-czi: I'm hesitant to create StackGetter, StackSetter, and StackDeleter when they're not strictly needed at this point. But I'm renaming StackIface to StackManager

jakeyheath (approved):

api/pkg/cmd/stack.go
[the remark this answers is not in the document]
hspitzley-czi: `MakeStack` runs at application boot time but the switch needs to execute per request

PR #1106, opened 2023-02-08
feat: reject requests from old tf provider versions

hspitzley-czi:


alexlokshin-czi (approved):

api/pkg/request/version_check_handler.go
```diff
@@ -14,7 +14,8 @@ var (
 
 func init() {
 	MinimumVersions = map[string]string{
-		"happy-cli": "0.52.0",
+		"happy-cli":      "0.52.0",
+		"happy-provider": "0.52.0",
```
jakeyheath, line 18: is this the user-agent? or will this need to reflect the mirror's repo name `terraform-provider-happy`?

jakeyheath (approved):
see comment before merging

api/pkg/request/version_check_handler.go
hspitzley-czi: it's the user agent, in this case it's defined here https://github.com/chanzuckerberg/happy/blob/2723b96edef3605e50c7f32e7027c991b67bf430/terraform/provider/pkg/provider/provider.go#L237

PR #1108, opened 2023-02-08
feat!: inject happy config to stacks

jakeyheath:


alexlokshin-czi (approved):

terraform/modules/happy-stack-eks/happy_config.tf
```diff
@@ -0,0 +1,9 @@
+data "happy_resolved_app_configs" "configs" {
+  app_name    = local.secret["tags"]["project"]
```
hspitzley-czi, line 2: is there a reason you use `local.secret["tags"]["project"]` instead of `var.app_name`?

terraform/modules/happy-service-ecs/variables.tf
```diff
@@ -125,7 +125,7 @@ variable "launch_type" {
 }
 
 variable "additional_env_vars" {
-  type        = list(object({ name : string, value : string }))
+  type        = map(string)
```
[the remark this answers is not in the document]
jakeyheath: good point

terraform/modules/happy-stack-eks/happy_config.tf
jakeyheath: I don't think var.app_name exists. We could explicitly require the user provide the variable in their stack, but I think it might be easier to infer it from the environment. Maybe this is a bad idea?

hspitzley-czi: oh right, it exists for happy-stack-ecs but not happy-stack-eks, seems like if one requires it they both should 🤷

jakeyheath: or both shouldn't. I'm tempted to take the interface from EKS and adopt it to ECS since EKS has users already and no one is on the ECS one yet

jakeyheath: @alexlokshin-czi will probably also want to pass it in though because he's thinking about multi-app clusters.

terraform/modules/happy-stack-eks/happy_config.tf
```diff
@@ -1,5 +1,5 @@
 data "happy_resolved_app_configs" "configs" {
-  app_name    = local.secret["tags"]["project"]
+  app_name    = len(var.app_name) == 0 ? local.secret["tags"]["project"]: var.app_name
```
jakeyheath, line 2: @hspitzley-czi how's this for a compromise?

hspitzley-czi: looks good to me

hspitzley-czi (approved):

PR #1138, opened 2023-02-13
fix: Happy version command stopped working

alexlokshin-czi:


jakeyheath (approved):

hspitzley-czi (changes requested):
Should this util be moved back to cli instead of a shared util since it's specific to the cli package?

alexlokshin-czi:
> Should this util be moved back to cli instead of a shared util since it's specific to the cli package?

We synchronize the release version across the board; instead, I believe the substitution needs to be implemented for API as well.

hspitzley-czi (approved):

PR #1144, opened 2023-02-14
feat: allow oidc providers to be configured by env variable

jakeyheath:
This PR allows the deployer of hapi to configure OIDC providers from a static configuration file and through an environment variable. The environment variable should take the form of:

~~~
HAPI_AUTH_OIDC_PROVIDERS: "issuer1|clientid1,issuer2|clientid2"
~~~

The contents of "HAPI_AUTH_OIDC_PROVIDERS" will be parsed and appended to the set of the OIDC providers created by the normal app-config.yml. This is useful for automations that will need to add to this list over time and redeploy the application through API calls rather than making PRs to this repo to update the repository.


api/pkg/setup/setup.go
```diff
@@ -162,12 +184,26 @@ func GetConfiguration() (*Configuration, error) {
 		}
 	}
 
-	cfg := &Configuration{}
-	err = vpr.Unmarshal(cfg)
+	// unique case where we want to be able to configure static OIDC providers
```
jakeyheath, line 187: @hspitzley-czi I got this working, although is the behavior you'd expect? Or should I just allow the environment variable HAPI_OIDCPROVIDERS override whatever is the configuration file? I feel like having a magic environment variable the overwrites stuff seems odds, but I'm not sure. I stared at viper too long today

jakeyheath: If you can think of a more general purpose way of doing this rather than only for the auth providers, that might be helpful down the line. Like something we could also use for other fields in this config file.

jakeyheath: 🤔 I think I am going to keep this for now, but something doesn't feel correct to me. Just haven't identified it yet

hspitzley-czi (approved):

PR #1149, opened 2023-02-15
feat: allow users to create bypasses for their OIDC

jakeyheath:
This PR allows users to create bypass rules to their OIDC protected stacks. This is handy for things like healthchecks and OPTIONS requests where cookies are not always present or the request is not always initiated by the user in a browser. This adds a good amount of complexity to these modules so I have added lots of validation rules and I would appreciate some more stringent pairs of eyes to look it over and make sure that if YOU were reading this module, you could still safely follow along. Here is an example of what something would look like if you failed to put the input variables in the wrong way:

Screenshot 2023-02-15 at 3 15 28 PM

It produces the something like the following AWS Load balancer set of rules:

Screenshot 2023-02-15 at 3 18 28 PM


terraform/modules/happy-ingress-eks/main.tf
```diff
@@ -1,52 +1,114 @@
...
   }
+
+  redirect_action_name = "redirect"
   ingress_tls_annotations = {
-    "alb.ingress.kubernetes.io/actions.redirect" = <<EOT
-        {"Type": "redirect", "RedirectConfig": {"Protocol": "HTTPS", "Port": "443", "StatusCode": "HTTP_301"}}
-      EOT
-    "alb.ingress.kubernetes.io/certificate-arn"  = var.certificate_arn
-    "alb.ingress.kubernetes.io/ssl-policy"       = "ELBSecurityPolicy-TLS-1-2-2017-01"
+    "alb.ingress.kubernetes.io/actions.${local.redirect_action_name}" = jsonencode({
```
jakeyheath, line 21: just for readability

terraform/modules/happy-ingress-eks/main.tf
```diff
@@ -1,52 +1,114 @@
...
-  listen_ports_base = [{ "HTTP" : 80 }]
-  listen_ports_tls  = [merge(local.listen_ports_base[0], { "HTTPS" : 443 })]
   ingress_base_annotations = {
+    "kubernetes.io/ingress.class"                    = "alb"
     "alb.ingress.kubernetes.io/backend-protocol"     = "HTTP"
     "alb.ingress.kubernetes.io/healthcheck-path"     = var.health_check_path
     "alb.ingress.kubernetes.io/healthcheck-protocol" = "HTTP"
-    # All ingresses are "internet-facing" so we need them all to listen on TLS
-    "alb.ingress.kubernetes.io/listen-ports" = jsonencode(local.listen_ports_tls)
+    "alb.ingress.kubernetes.io/listen-ports"         = jsonencode([{ HTTPS = 443 }, { HTTP = 80 }])
```
jakeyheath, line 7: readability

terraform/modules/happy-ingress-eks/main.tf
```diff
@@ -1,52 +1,114 @@
...
   )
+
+  // the length of bypasses should never be bigger than the priority due to how we do the spread priority
+  // in happy-stack-eks. We also have a validation on the input variable to ensure this.
+  priority_spread = range(var.routing.priority - length(var.routing.bypasses), var.routing.priority)
+  routing_keys    = keys(var.routing.bypasses)
+  ingress_bypass_annotations = ({
+    for k, v in zipmap(local.routing_keys, local.priority_spread) : k =>
+    merge(
+      local.ingress_base_annotations,
```
jakeyheath, line 52: `var.routing.priority` is the base routing number and the priority_spread is the set of overrides to use. Here `v` is assigned a number smaller than `var.routing.priority`

alexlokshin-czi (approved):

PR #1174, opened 2023-02-17
feat: Happy application autoscaling support

alexlokshin-czi:


terraform/modules/happy-service-eks/main.tf
```diff
@@ -210,3 +210,23 @@ module "ingress" {
   tags_string     = local.tags_string
   routing         = var.routing
 }
+
+resource "kubernetes_horizontal_pod_autoscaler_v1" "hpa" {
```
jakeyheath, line 214: ooooo

terraform/modules/happy-service-eks/main.tf
```diff
@@ -210,3 +210,23 @@ module "ingress" {
...
+  metadata {
+    name      = var.routing.service_name
+    namespace = var.k8s_namespace
+  }
+
+  spec {
+    max_replicas = var.max_count
+    min_replicas = var.desired_count
+
+    target_cpu_utilization_percentage = 80
```
jakeyheath, line 224: are there other metrics to scale on? is CPU util pretty common indicator to scale up?

jakeyheath (approved):

terraform/modules/happy-service-eks/main.tf
alexlokshin-czi: :)

terraform/modules/happy-service-eks/main.tf
alexlokshin-czi: HPAs support CPU and memory; `kubernetes_horizontal_pod_autoscaler_v1` only allows us to specify cpu if we don't want to use the standard behavior (which accounts for both heuristically).

PR #1217, opened 2023-02-23
feat: List of all AWS infra associated with a k8s happy stack

alexlokshin-czi:
Addresses [CCIE-916]. Retrieves stack-level resources from terraform state, as well as k8s resources that have been created by them (Route 53 entries, ALBs). Output of `happy resources stackname` looks like this:

Screenshot 2023-02-23 at 10 58 31 AM


cli/cmd/resources.go
```diff
@@ -0,0 +1,84 @@
...
+		stackName := args[0]
+
+		bootstrapConfig, err := config.NewBootstrapConfig(cmd)
+		if err != nil {
+			return err
+		}
+		happyConfig, err := config.NewHappyConfig(bootstrapConfig)
+		if err != nil {
+			return err
+		}
```
hspitzley-czi, line 38: There's a helper for this:
```suggestion
		happyConfig, err := config.GetHappyConfigForCmd(cmd)
		if err != nil {
			return err
		}
```

hspitzley-czi (approved):

PR #1232, opened 2023-02-23
feat!: refactor update/clean; autocreate ECR

jakeyheath:
This PR cleans up happy create a little bit and isolates the three pieces of what happens when a create happens. This allows us to move around some of the components so we can do a targeted ECR apply to move that code into the stack. This is nice because in the past, when happy users want to create new services in their docker-compose/happy deployment, they first needed to make a commit to a durable infrastructure to add an ECR and make sure the name was exactly right. They'd do this for all their environments. 

Now, if the feature flag is enabled, happy will do a targeted deploy that will create the ECR before building and pushing. Each stack has its own ECR and it will be destroyed when the stack is destroyed. Users are no longer required to update their durable infrastructure when making new services.


alexlokshin-czi (approved):

cli/cmd/create.go
```diff
@@ -48,259 +33,69 @@ var createCmd = &cobra.Command{
...
-		buildOpts = append(buildOpts, artifact_builder.BuildSlice(slice))
-		builderConfig.WithProfile(slice.Profile)
-	}
-	ab := artifact_builder.NewArtifactBuilder(isDryRun).WithConfig(builderConfig).WithBackend(backend)
-
-	workspaceRepo := createWorkspaceRepo(isDryRun, backend)
-
-	err = verifyTFEBacklog(ctx, workspaceRepo)
+	happyConfig, stackService, artifactBuilder, tag, stackTags, awsBackend, err := initializeHappyClients(
+		cmd,
```
alexlokshin-czi, line 50: I feel like the number of arguments this method returns is excessive, there's a lot of implied complexity in this approach, it's better to wrap happyConfig, stackService, artifactBuilder, tag, stackTags, awsBackend into a separate struct and return thatstruct, err.

cli/cmd/create.go
```diff
@@ -48,259 +33,69 @@ var createCmd = &cobra.Command{
...
-	err = options.StackMeta.Update(ctx, targetBaseTag, options.StackTags, "", options.StackService)
-	if err != nil {
-		return errors.Wrap(err, "failed to update the stack meta")
-	}
-
-	stack, err := options.StackService.Add(ctx, options.StackName, isDryRun)
+	stackName := args[0]
+	err = validateHappyEnvironment(
+		ctx,
+		happyConfig,
```
alexlokshin-czi, line 63: Same as above goes for the input parameters -- if a number of them are used together, they should be aggregated.

cli/cmd/create.go
```diff
@@ -48,259 +33,69 @@ var createCmd = &cobra.Command{
...
-		}
-
-		if !proceed {
-			return err
-		}
-	}
-	return nil
+	// 2.) otherwise, update the existing stacks
+	return updateStack(
+		ctx,
```
alexlokshin-czi, line 89: Same as above, too many arguments.

cli/cmd/update.go
```diff
@@ -36,193 +35,138 @@ var updateCmd = &cobra.Command{
...
+	)
 	if err != nil {
-		return err
+		return errors.Wrap(err, "unable to initialize all the happy clients")
 	}
 
-	happyConfig, err := config.NewHappyConfig(bootstrapConfig)
+	stackName := args[0]
+	err = validateHappyEnvironment(
+		ctx,
```
alexlokshin-czi, line 60: Same as above.

cli/cmd/update.go
```diff
@@ -36,193 +35,138 @@ var updateCmd = &cobra.Command{
...
-		buildOpts = append(buildOpts, artifact_builder.BuildSlice(slice))
-		builderConfig.WithProfile(slice.Profile)
 	}
 
-	ab := artifact_builder.NewArtifactBuilder(isDryRun).WithBackend(backend).WithConfig(builderConfig)
-	workspaceRepo := createWorkspaceRepo(isDryRun, backend)
-	stackService := stackservice.NewStackService().WithBackend(backend).WithWorkspaceRepo(workspaceRepo)
+	// 2.) update the existing stacks
+	return updateStack(
+		ctx,
```
alexlokshin-czi, line 87: Same as above.

cli/cmd/create.go
jakeyheath: 👍🏻 ya agreed

cli/cmd/create.go
```diff
@@ -48,259 +37,109 @@ var createCmd = &cobra.Command{
...
-	// if creating tag, build and push images
-	if createTag {
-		// parity with update command
-		buildOpts = append(buildOpts, artifact_builder.WithTags(tag))
-		err = ab.BuildAndPush(ctx, buildOpts...)
+			return true, nil
+		})()
 		if err != nil {
-			return err
+			return errors.Wrap(err, "unable to get ECRs for service")
```
alexlokshin-czi, line 104: Might be useful to include the name of the service for which we cannot get the ECR

cli/pkg/artifact_builder/artifact_builder.go
```diff
@@ -33,14 +33,18 @@ type ArtifactBuilder struct {
 	backend  *backend.Backend
 	config   *BuilderConfig
 	Profiler *profiler.Profiler
-	tags     []string
+	Tags     []string
```
alexlokshin-czi, line 36: Do we need to export `Tags`? There's a `GetTags()` func. Two conflicting access methods.

cli/pkg/artifact_builder/artifact_builder.go
```diff
@@ -53,7 +57,7 @@ func (ab ArtifactBuilder) WithBackend(backend *backend.Backend) ArtifactBuilderI
 
 func (ab ArtifactBuilder) WithTags(tags []string) ArtifactBuilderIface {
 	if len(tags) > 0 {
-		ab.tags = tags
+		ab.Tags = tags
```
alexlokshin-czi, line 60: As you can see, `WithTags()` becomes redundant as `Tags` can be accessed directly. Or through `SetTags()`.

cli/pkg/artifact_builder/artifact_builder.go
```diff
@@ -75,6 +79,16 @@ func (ab ArtifactBuilder) CheckImageExists(ctx context.Context, tag string) (boo
 		return false, errors.Wrap(err, "artifact builder configuration is incomplete")
 	}
 	serviceRegistries := ab.backend.Conf().GetServiceRegistries()
+	// backward compatible way of overriding the new ECR locations
+	// if they exist. The new ECRs will look like <stackname>-<env>-<servicename>
+	// if users haven't switched to the latest stack TF module, this will return nothing
+	stackECRS, err := ab.GetECRsForServices(ctx)
+	if err != nil {
+		log.Error(err)
```
alexlokshin-czi, line 87: Is this really an error if it's ignored?

cli/pkg/artifact_builder/artifact_builder.go
```diff
@@ -219,23 +233,58 @@ func (ab ArtifactBuilder) RegistryLogin(ctx context.Context) error {
...
+			break
+		}
+		nextToken = r.NextToken
+	}
+	ecrs := map[string]*config.RegistryConfig{}
+	for _, service := range ab.backend.Conf().HappyConfig.GetServices() {
+		for _, repo := range repos {
+			if repo.RepositoryName != nil &&
+				repo.RepositoryUri != nil &&
+				*repo.RepositoryName == fmt.Sprintf("%s-%s-%s", ab.config.StackName, ab.config.env, service) {
```
alexlokshin-czi, line 280: I propose we use `/` as a separator character for readability. We can have conflicting repo names, as we often use dashes in our naming. `jheath-prod-prod-frontend` could be interpreted in several different ways, why `jheath-prod/prod/frontend` won't. I believe we already disallow `/` as a part of the stack name, so no duplicity there.

jakeyheath: I could also run the terraform targeted apply each time and have terraform manage that.

jakeyheath: o actually, I think I can use the state of the targeted apply to get these. I don't think I need to make an API call.

jakeyheath:
> 

I'm not actually sure :/ . Most of the changes are moving a few functions around and consolidating. So our normal tests should work as they did before. the behavior of the ECR creation is new and maybe I could write some tests that show case that.

PR #1233, opened 2023-02-24
feat: prevent changes in dynamic tags

jakeyheath:
The tags were changing on every plan because I think the order of `var.tags` was producing a randomly sorted list. Adding a sort argument will prevent the plan from producing so many changes.


alexlokshin-czi (approved):

alexlokshin-czi:
Nice!!!

PR #1296, opened 2023-03-07
feat: Include a specific message in creator workspace runs and add happy version number to the message

alexlokshin-czi:


jakeyheath (approved):
sick

PR #1315, opened 2023-03-10
feat: 'happy logs' integration with Cloudwatch Insights

alexlokshin-czi:
This improvement gives you access to Cloudwatch Insights straight from Happy. If you're running Happy CLI in a CICD mode (`--detached`), we will drop a message into the output with the Cloudwatch insights link to your logs; when running interactively, we will prompt the browser window to open, and navigate straight to your logs to query and visualize. Supports both EKS and ECS computes.
Screenshot 2023-03-10 at 4 47 11 PM
Screenshot 2023-03-10 at 4 48 14 PM


cli/cmd/root.go
```diff
@@ -59,39 +59,42 @@ var rootCmd = &cobra.Command{
...
+		}
+
+		err = util.ValidateEnvironment(cmd.Context())
 		return errors.Wrap(err, "local environment is misconfigured")
 	},
 }
 
 // Execute executes the command
 func Execute() error {
-	detached, err := rootCmd.Flags().GetBool(flagDetached)
```
alexlokshin-czi, line 69: Had to move this code around because original implementation plain didn't work.

cli/pkg/backend/aws/k8s_compute_backend.go
```diff
@@ -129,7 +129,30 @@ func (k8s *K8SComputeBackend) PrintLogs(ctx context.Context, stackName string, s
 			logrus.Error(err.Error())
 		}
 	}
-	return nil
+
+	if k8s.KubeConfig.AuthMethod != kube.AuthMethodEKS {
+		return nil
+	}
+
+	expression := fmt.Sprintf(`fields @timestamp, log
```
alexlokshin-czi, line 137: When all stacks receive new annotations, will select by kubernetes lables, like `part-of` or `component`. For now, assuming pods are named `stackname-servicename`.

cli/cmd/init.go
```diff
@@ -43,7 +42,7 @@ func makeHappyClient(cmd *cobra.Command, sliceName, stackName, tag string, creat
 	if err != nil {
 		return nil, err
 	}
-	builderConfig := artifact_builder.
+	builderConfig := ab.
```
netvisao, line 45: Where is ab coming from?, names should be more descriptive than 1 letter, improves readability.

cli/pkg/backend/aws/backend.go
```diff
@@ -281,7 +285,7 @@ func (b *Backend) PrintLogs(ctx context.Context, stackName string, serviceName s
 	return b.ComputeBackend.PrintLogs(ctx, stackName, serviceName, opts...)
 }
 
-func (b *Backend) RunTask(ctx context.Context, taskDefArn string, launchType config.LaunchType) error {
+func (b *Backend) RunTask(ctx context.Context, taskDefArn string, launchType util.LaunchType) error {
```
netvisao, line 288: we need the variable names to be more descriptive than a letter

netvisao (changes requested):

cli/cmd/init.go
alexlokshin-czi: It's a package name alias declared at the top of the file.

cli/pkg/backend/aws/backend.go
alexlokshin-czi: This construct refers to a binding type struct that will have these functions assigned to. Happy throughout the codebase has shortened var names for the grouping struct as it is not a traditional variable. It's not the name, but the type of the grouping struct that matters, and it is descriptive (`Backend`).

netvisao: That only works for functions with limited lines of code; using cryptic names will affect readability as they grow.  Case in point with the ab situation, where I had to expand the file to figure out what ab is.
https://readlearncode.com/code-and-stuff/clean-code-variables/

netvisao (changes requested):

cli/pkg/backend/aws/backend.go
alexlokshin-czi: `ab` is not a variable; it's a pretty standard convention in golang to shorten conflicting or lengthy package names.

alexlokshin-czi: Here's an example from kubernetes code: https://github.com/kubernetes/kubernetes/blob/fcfe5dfc2177320833caa451b401f2d1dc5668e7/pkg/securitycontext/util.go#L20

alexlokshin-czi: As this code is written by convention, and is consistent across packages, we are not sacrificing readability. Go is a language of limited verbosity, unlike Java.

cli/cmd/root.go
```diff
@@ -59,39 +59,41 @@ var rootCmd = &cobra.Command{
 			color.NoColor = noColor
 		}
 
-		err = util.ValidateEnvironment(context.Background())
+		detached, err := cmd.Flags().GetBool(flagDetached)
```
jakeyheath, line 62: is this related to validating if the image exists? I don't know if this is the right place for it. Also, if you move this code, can you pass in the flag values rather than using the global values.

cli/pkg/backend/aws/backend.go
```diff
@@ -296,3 +301,35 @@ func (b *Backend) GetEvents(ctx context.Context, stackName string, services []st
...
+			}
+			logrus.Info("Opening Browser window to query cloudwatch insights.")
+			err = browser.OpenURL(cloudwatchLink)
+			if err != nil {
+				return errors.Wrap(err, "To our dismay, we were unable open up a browser window to query cloudwatch insights.")
+			}
+			logrus.Info("Select the desired time frame, and click 'Run Query' to query the logs.")
+			return nil
+		}
+		logrus.Info("****************************************************************************************")
```
jakeyheath, line 327: would you mind showing what this looks like in the terminal and put it in the PR description

cli/cmd/root.go
alexlokshin-czi: This is the right place for this code; none of these flags are accessible in `Execute()` context.

cli/pkg/backend/aws/backend.go
alexlokshin-czi: Done.

cli/pkg/stack_mgr/stack_service.go
```diff
@@ -210,7 +209,7 @@ func (s *StackService) removeFromStacklist(ctx context.Context, stackName string
 	return s.writeStacklist(ctx, stackNamesList)
 }
 
-func (s *StackService) Add(ctx context.Context, stackName string, dryRun bool, options ...workspace_repo.TFERunOption) (*Stack, error) {
+func (s *StackService) Add(ctx context.Context, stackName string, dryRun bool, options ...workspacerepo.TFERunOption) (*Stack, error) {
```
jakeyheath, line 212: if we are adding run options to all these methods, there is a DryRun run option, so we can probably replace the dryRun argument in all of them with a DryRun TFERunOption

cli/cmd/root.go
jakeyheath: o duh nevermind. I misread where this was

jakeyheath (approved):

cli/pkg/stack_mgr/stack_service.go
alexlokshin-czi: Let's do it in a separate PR -- this change is too intrusive.

netvisao (approved):

PR #1355, opened 2023-03-14
feat: CCIE-900 Automatically check whether Happy is up to date

dspadea:
- Add Release.Equal() to compare releases
- Add Function to fetch latest available version from hapi
- Do automated version check after command executes, and warn if Happy is not current
- Expose available-version command to allow on-demand fetching of current version number of Happy


cli/cmd/root.go
```diff
@@ -107,5 +107,24 @@ func Execute() error {
...
+	outdated, cliVersion, latestAvailableVersion, err := IsHappyOutdated(cmd)
+
+	if err != nil {
+		log.Errorf("Error checking for latest available version number: %v", err)
+		return
+	}
+
+	if outdated {
+		log.Warnf("This copy of Happy CLI is not the latest availble. Please update at your earliest convenience.")
+		log.Warnf("CLI version: %s  Latest available version: %s", cliVersion.Version, latestAvailableVersion.Version)
```
jakeyheath, line 127: Can we print the command used to update?

`brew upgrade happy`

dspadea: Sure! Good idea!

dspadea: Now reads:

```
[WARNING]: This copy of Happy CLI is not the latest available. CLI version: 999.12.34-SNAPSHOT-26b60f8  Latest available version: v0.64.0

[WARNING]: To update on Mac, run:  brew upgrade happy
```

cli/cmd/version.go
```diff
@@ -22,3 +26,49 @@ var versionCmd = &cobra.Command{
...
+
+	api := hapi.MakeApiClient(happyConfig)
+	result := model.HealthResponse{}
+	err = api.GetParsed("/health", "", &result)
+	if err != nil {
+		return nil, err
+	}
+
+	return &util.Release{
+		Version: result.Version,
```
jakeyheath, line 61: Also, just to confirm. The version that comes back from the API, this is the CLI version? Not the API version right?

jakeyheath: jk, I just checked it out. These should be the same

dspadea: Good point, and thanks for checking!

PR #1359, opened 2023-03-15
chore: create dynamo table for EKS

jakeyheath:
This PR adds a dynamo table that should be used for stacklist locking. The locking is used by happy to make sure that the stacklist is thread safe.


terraform/modules/happy-env-eks/dynamo.tf
```diff
@@ -0,0 +1,39 @@
...
+  write_capacity = 5
+  hash_key       = "key"
+
+  attribute {
+    name = "key"
+    type = "S"
+  }
+
+  server_side_encryption {
+    enabled = true
```
alexlokshin-czi, line 14: Do we need encryption for the lock table?

jakeyheath: not sure I copied to from ECR

PR #1371, opened 2023-03-15
feat: CCIE-858: lock the version of happy cli in a repo like fogg

dspadea:
Implement (currently optional) Happy version locking in repos.

Scenarios:

- Fail check silently and proceed if we can't get the latest available version (generally if we are outside a project root)
- Allow for version, available-version, and set-lock commands to exclude locked version check
- Allow for legacy behavior if no lockfile is present (i.e. proceed with any version)
- Refuse to run for most commands if lockfile is present and version does not match

### DIFF
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"context"
+	"fmt"
 	"net/url"
 	"time"
 
@@ -24,6 +25,8 @@ const (
 
 var OutputFormat string = "text"
 var Interactive bool = true
+
+var excludeVersionCheckCmds = map[string]interface{}{"version": nil, "set-lock": nil}
 
 func init() {
 	rootCmd.PersistentFlags().BoolP(flagVerbose, "v", false, "Use this to enable verbose mode")
@@ -82,12 +85,17 @@ var rootCmd = &cobra.Command{
 			}
 		}
 
-		if versionMatch, cliVersion, lockedVersion, err := VerifyHappyIsLockedVersion(cmd); err != nil {
-			return errors.Wrap(err, "Unable to verify locked Happy version")
-		} else {
-			if !versionMatch {
-				return errors.Errorf("Installed Happy version (%s) does not match locked version in .happy/version.lock (%s)", cliVersion, lockedVersion)
+		fmt.Printf("Current command: %s\n", cmd.CalledAs())
+		if _, present := excludeVersionCheckCmds[cmd.CalledAs()]; !present {
+			if versionMatch, cliVersion, lockedVersion, err := VerifyHappyIsLockedVersion(cmd); err != nil {
+				return errors.Wrap(err, "Unable to verify locked Happy version")
+			} else {
+				if !versionMatch {
+					return errors.Errorf("Installed Happy version (%s) does not match locked version in .happy/version.lock (%s)", cliVersion, lockedVersion)
+				}
 			}
+		} else {
+			log.Debug("Skipping locked version check")
 		}
 
 		err = util.ValidateEnvironment(cmd.Context())
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -85,6 +85,10 @@ var rootCmd = &cobra.Command{
 		err = util.ValidateEnvironment(cmd.Context())
 		return errors.Wrap(err, "local environment is misconfigured")
 	},
+
+	PersistentPostRun: func(cmd *cobra.Command, args []string) {
+		WarnIfHappyOutdated(cmd)
+	},
 }
 
 // Execute executes the command
@@ -107,8 +111,6 @@ func Execute() error {
 			log.Warn(warning)
 		}
 	}
-
-	WarnIfHappyOutdated(rootCmd)
 
 	return nil
 }
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -114,17 +114,3 @@ func Execute() error {
 
 	return nil
 }
-
-func WarnIfHappyOutdated(cmd *cobra.Command) {
-	outdated, cliVersion, latestAvailableVersion, err := IsHappyOutdated(cmd)
-
-	if err != nil {
-		log.Errorf("Error checking for latest available version number: %v", err)
-		return
-	}
-
-	if outdated {
-		log.Warnf("This copy of Happy CLI is not the latest available. CLI version: %s  Latest available version: %s\n", cliVersion.Version, latestAvailableVersion.Version)
-		log.Warn("To update on Mac, run:  brew upgrade happy")
-	}
-}
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -82,7 +82,8 @@ var rootCmd = &cobra.Command{
 			}
 		}
 
-		if err = CheckLockedHappyVersion(cmd); err != nil {
+		err = CheckLockedHappyVersion(cmd)
+		if err != nil {
 			return err
 		}
 
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -87,9 +87,7 @@ var rootCmd = &cobra.Command{
 			return errors.Wrap(err, "Unable to verify locked Happy version")
 		} else {
 			if !versionMatch {
-				return errors.New(
-					fmt.Sprintf("Installed Happy version (%s) does not match locked version in .happy/version.lock (%s)", cliVersion, lockedVersion),
-				)
+				return errors.Errorf("Installed Happy version (%s) does not match locked version in .happy/version.lock (%s)", cliVersion, lockedVersion)
 			}
 		}
 
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -82,6 +82,14 @@ var rootCmd = &cobra.Command{
 			}
 		}
 
+		if versionMatch, err := VerifyHappyIsLockedVersion(cmd); err != nil {
+			return errors.Wrap(err, "Unable to verify locked Happy version")
+		} else {
+			if !versionMatch {
+				return errors.New("Installed Happy version does not match locked version in .happy/version.lock")
+			}
+		}
+
 		err = util.ValidateEnvironment(cmd.Context())
 		return errors.Wrap(err, "local environment is misconfigured")
 	},
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -2,7 +2,6 @@ package cmd
 
 import (
 	"context"
-	"fmt"
 	"net/url"
 	"time"
 
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -2,7 +2,6 @@ package cmd
 
 import (
 	"context"
-	"fmt"
 	"net/url"
 	"time"
 
@@ -25,8 +24,6 @@ const (
 
 var OutputFormat string = "text"
 var Interactive bool = true
-
-var excludeVersionCheckCmds = map[string]interface{}{"version": nil, "set-lock": nil}
 
 func init() {
 	rootCmd.PersistentFlags().BoolP(flagVerbose, "v", false, "Use this to enable verbose mode")
@@ -85,17 +82,8 @@ var rootCmd = &cobra.Command{
 			}
 		}
 
-		fmt.Printf("Current command: %s\n", cmd.CalledAs())
-		if _, present := excludeVersionCheckCmds[cmd.CalledAs()]; !present {
-			if versionMatch, cliVersion, lockedVersion, err := VerifyHappyIsLockedVersion(cmd); err != nil {
-				return errors.Wrap(err, "Unable to verify locked Happy version")
-			} else {
-				if !versionMatch {
-					return errors.Errorf("Installed Happy version (%s) does not match locked version in .happy/version.lock (%s)", cliVersion, lockedVersion)
-				}
-			}
-		} else {
-			log.Debug("Skipping locked version check")
+		if err = CheckLockedHappyVersion(cmd); err != nil {
+			return err
 		}
 
 		err = util.ValidateEnvironment(cmd.Context())
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"context"
+	"fmt"
 	"net/url"
 	"time"
 
@@ -82,11 +83,13 @@ var rootCmd = &cobra.Command{
 			}
 		}
 
-		if versionMatch, err := VerifyHappyIsLockedVersion(cmd); err != nil {
+		if versionMatch, cliVersion, lockedVersion, err := VerifyHappyIsLockedVersion(cmd); err != nil {
 			return errors.Wrap(err, "Unable to verify locked Happy version")
 		} else {
 			if !versionMatch {
-				return errors.New("Installed Happy version does not match locked version in .happy/version.lock")
+				return errors.New(
+					fmt.Sprintf("Installed Happy version (%s) does not match locked version in .happy/version.lock (%s)", cliVersion, lockedVersion),
+				)
 			}
 		}
 
--- a/cli/cmd/version.go
+++ b/cli/cmd/version.go
@@ -186,16 +186,18 @@ func VerifyHappyIsLockedVersion(cmd *cobra.Command) (bool, string, string, error
 
 func CheckLockedHappyVersion(cmd *cobra.Command) error {
 
-	excludeVersionCheckCmds := map[string]interface{}{
+	excludeLockedVersionCheckCmds := map[string]interface{}{
 		"version":           nil,
 		"set-lock":          nil,
 		"available-version": nil,
 	}
 
 	log.Debugf("Current command: %s\n", cmd.CalledAs())
-	if _, present := excludeVersionCheckCmds[cmd.CalledAs()]; !present {
+	if _, present := excludeLockedVersionCheckCmds[cmd.CalledAs()]; !present {
 		if versionMatch, cliVersion, lockedVersion, err := VerifyHappyIsLockedVersion(cmd); err != nil {
-			return errors.Wrap(err, "Unable to verify locked Happy version")
+			// This is generally going to be because we are outside of a project root.
+			log.Debugf("Unable to verify locked Happy version: %s", err)
+			return nil
 		} else {
 			if !versionMatch {
 				return errors.Errorf("Installed Happy version (%s) does not match locked version in .happy/version.lock (%s)", cliVersion, lockedVersion)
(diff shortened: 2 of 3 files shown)


cli/cmd/root.go
```diff
@@ -82,9 +83,23 @@ var rootCmd = &cobra.Command{
 			}
 		}
 
+		if versionMatch, cliVersion, lockedVersion, err := VerifyHappyIsLockedVersion(cmd); err != nil {
+			return errors.Wrap(err, "Unable to verify locked Happy version")
+		} else {
+			if !versionMatch {
+				return errors.New(
+					fmt.Sprintf("Installed Happy version (%s) does not match locked version in .happy/version.lock (%s)", cliVersion, lockedVersion),
+				)
```
jakeyheath, line 92:
```suggestion
				return errors.Error("Installed Happy version (%s) does not match locked version in .happy/version.lock (%s)", cliVersion, lockedVersion)
```

jakeyheath: let's also provide the command to get that particular version (I always forget the brew commands)

cli/cmd/version.go
```diff
@@ -39,7 +43,24 @@ var availableVersionCmd = &cobra.Command{
...
+
+var lockHappyVersionCmd = &cobra.Command{
+	Use:          "lock",
+	Short:        "Create a .happy/version.lock file",
+	Long:         "Create a .happy/version.lock file in project root to specify which version of Happy should be used with this project. This will overwrite any existing version.lock file.",
+	SilenceUsage: true,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		versionFile, version, err := CreateHappyVersionFile(cmd)
+		if err != nil {
+			fmt.Fprint(cmd.Parent().ErrOrStderr(), err)
```
jakeyheath, line 59:
```suggestion
			log.Debug(cmd.Parent().ErrOrStderr(), err)
```

unless this was just for testing, in which case ignore this comment and remove before its finished

cli/cmd/version.go
```diff
@@ -72,3 +93,89 @@ func IsHappyOutdated(cmd *cobra.Command) (bool, *util.Release, *util.Release, er
...
+	currentVersion := util.GetVersion()
+	projectRoot := happyConfig.GetProjectRoot()
+
+	requestedVersion, _ := cmd.Flags().GetString("version")
+
+	if requestedVersion == "" {
+		requestedVersion = currentVersion.Version
+	}
+
+	versionFile := config.NewHappyVersionLockFile(projectRoot)
```
jakeyheath, line 128: might be easier to write a test for this function if these were arguments to CreateHappyVersionFile instead of the cmd. So something like `CreateHappyVersionFile(requestedVersion, currentVersion projectRoot string, happyConfig *HappyConfig)`

cli/pkg/config/happy_version_lock.go
```diff
@@ -0,0 +1,112 @@
...
+
+	return v.HappyVersion, nil
+}
+
+func (v *HappyVersionLockFile) GetPath() (string, error) {
+	if v.path == "" {
+		return "", errors.New("Path is not set")
+	}
+	return v.path, nil
+}
```
jakeyheath, line 112: generally, I would avoid the getter/setter pattern and expose the attribute. I would focus on error handling at the constructor `NewHappyVersionLockFile` instead

cli/pkg/config/happy_version_lock.go
```diff
@@ -0,0 +1,112 @@
...
+
+	happyVersionFile.WriteString(string(contents))
+	happyVersionFile.Close()
+
+	return nil
+}
+
+func calcHappyVersionPath(projectRoot string) string {
+	versionFilePath := filepath.Join(projectRoot, ".happy", "version.lock")
+	return versionFilePath
```
jakeyheath, line 84:
```suggestion
	return filepath.Join(projectRoot, ".happy", "version.lock")
```

cli/pkg/config/happy_version_lock.go
dspadea: Ok - Updated accordingly. Thanks!

cli/cmd/version.go
```diff
@@ -67,8 +88,123 @@ func IsHappyOutdated(cmd *cobra.Command) (bool, *util.Release, *util.Release, er
...
+		if versionMatch, cliVersion, lockedVersion, err := VerifyHappyIsLockedVersion(cmd); err != nil {
+			// This is generally going to be because we are outside of a project root.
+			log.Debugf("Unable to verify locked Happy version: %s", err)
+			return nil
+		} else {
+			if !versionMatch {
+				return errors.Errorf("Installed Happy version (%s) does not match locked version in .happy/version.lock (%s)", cliVersion, lockedVersion)
+			}
+		}
+	} else {
```
jakeyheath, line 206: I don't think this else is necessary

cli/cmd/version.go
```diff
@@ -67,8 +88,123 @@ func IsHappyOutdated(cmd *cobra.Command) (bool, *util.Release, *util.Release, er
...
+		log.Debug("Skipping locked version check")
+	}
```
jakeyheath, line 208:
```suggestion
	}
	log.Debug("Skipping locked version check")
```
I don't think this else is necessary

cli/cmd/root.go
```diff
@@ -82,9 +82,17 @@ var rootCmd = &cobra.Command{
 			}
 		}
 
+		if err = CheckLockedHappyVersion(cmd); err != nil {
+			return err
+		}
```
jakeyheath, line 87:
```suggestion
		err = CheckLockedHappyVersion(cmd)
		if err != nil {
			return err
		}
```
I think this is more conventional for this repo

cli/cmd/version.go
dspadea: Ah, right, because it returns otherwise in the if block. Good catch.

cli/cmd/version.go
```diff
@@ -67,8 +88,122 @@ func IsHappyOutdated(cmd *cobra.Command) (bool, *util.Release, *util.Release, er
...
+	if _, present := excludeLockedVersionCheckCmds[cmd.CalledAs()]; !present {
+		if versionMatch, cliVersion, lockedVersion, err := VerifyHappyIsLockedVersion(cmd); err != nil {
+			// This is generally going to be because we are outside of a project root.
+			log.Debugf("Unable to verify locked Happy version: %s", err)
+			return nil
+		} else {
+			if !versionMatch {
+				return errors.Errorf("Installed Happy version (%s) does not match locked version in .happy/version.lock (%s)", cliVersion, lockedVersion)
+			}
+		}
```
jakeyheath, line 205:
```suggestion
		} 
			if !versionMatch {
				return errors.Errorf("Installed Happy version (%s) does not match locked version in .happy/version.lock (%s)", cliVersion, lockedVersion)
			
		}
```
I don't think this else is necessary

cli/cmd/version.go
```diff
@@ -67,8 +88,122 @@ func IsHappyOutdated(cmd *cobra.Command) (bool, *util.Release, *util.Release, er
...
+
+	excludeLockedVersionCheckCmds := map[string]interface{}{
+		"version":           nil,
+		"set-lock":          nil,
+		"available-version": nil,
+	}
+
+	log.Debugf("Current command: %s\n", cmd.CalledAs())
+	if _, present := excludeLockedVersionCheckCmds[cmd.CalledAs()]; !present {
+		if versionMatch, cliVersion, lockedVersion, err := VerifyHappyIsLockedVersion(cmd); err != nil {
```
jakeyheath, line 197:
```suggestion
		versionMatch, cliVersion, lockedVersion, err := VerifyHappyIsLockedVersion(cmd)
		if err != nil {
```

more conventional for this repo

jakeyheath: I'm trying to find the style guide docs, but generally, I try to avoid as many nested things as possible. so if I see more then 2 nests, its a red flag for me to clean up

cli/pkg/config/happy_version_lock.go
```diff
@@ -0,0 +1,92 @@
...
+
+	if requiredVersion == "" {
+		return nil, errors.New("No requiredVersion specified")
+	}
+
+	path := calcHappyVersionPath(projectRoot)
+
+	return &HappyVersionLockFile{
+		HappyVersion: requiredVersion,
+		Path:         path,
```
jakeyheath, line 32: Did this get fixed?

cli/pkg/config/happy_version_lock.go
```diff
@@ -0,0 +1,92 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	hvlf := HappyVersionLockFile{}
+
+	err = json.Unmarshal(contents, &hvlf)
+	if err != nil {
+		return nil, err
+	}
```
jakeyheath, line 62: can you uses `errors.Wrap(err, "message")` on these errors?

cli/pkg/config/happy_version_lock.go
dspadea: Yeah. I excluded it in the struct definition.

cli/pkg/config/happy_version_lock.go
```diff
@@ -0,0 +1,92 @@
...
+	happyVersionFile, err := os.Create(v.Path)
+
+	if err != nil {
+		return errors.New(fmt.Sprintf("Could not create %s: %v", v.Path, err))
+	}
+
+	contents, err := json.MarshalIndent(&v, "", " ")
+	if err != nil {
+		return err
+	}
```
jakeyheath, line 82: same here. `errors.Wrap(err, "message")`or `errors.Wrapf(err, "message")`

cli/pkg/config/happy_version_lock.go
dspadea: Sure

cli/cmd/version.go
```diff
@@ -67,8 +88,123 @@ func IsHappyOutdated(cmd *cobra.Command) (bool, *util.Release, *util.Release, er
...
+
+	log.Debugf("Current command: %s\n", cmd.CalledAs())
+	if _, present := excludeLockedVersionCheckCmds[cmd.CalledAs()]; !present {
+		versionMatch, cliVersion, lockedVersion, err := VerifyHappyIsLockedVersion(cmd)
+		if err != nil {
+			// This is generally going to be because we are outside of a project root.
+			log.Debugf("Unable to verify locked Happy version: %s", err)
+			return nil
+		} 
+			if !versionMatch {
```
jakeyheath, line 203: can you resolve these lint-ci warnings?

cli/cmd/version.go
```diff
@@ -67,8 +88,123 @@ func IsHappyOutdated(cmd *cobra.Command) (bool, *util.Release, *util.Release, er
...
+func CheckLockedHappyVersion(cmd *cobra.Command) error {
+
+	excludeLockedVersionCheckCmds := map[string]interface{}{
+		"version":           nil,
+		"set-lock":          nil,
+		"available-version": nil,
+	}
+
+	log.Debugf("Current command: %s\n", cmd.CalledAs())
+	if _, present := excludeLockedVersionCheckCmds[cmd.CalledAs()]; !present {
```
jakeyheath, line 196:
```suggestion
	if _, present := excludeLockedVersionCheckCmds[cmd.CalledAs()]; present {
```
refactor to return fast if not present and have the main body of logic handle the case when the cmd is in the map

cli/cmd/version.go
```diff
@@ -67,8 +88,123 @@ func IsHappyOutdated(cmd *cobra.Command) (bool, *util.Release, *util.Release, er
...
+
+	err = versionFile.Save()
+	if err != nil {
+		return "", "", err
+	}
+
+	return versionFile.Path, versionFile.HappyVersion, nil
+}
+
+func VerifyHappyIsLockedVersion(cmd *cobra.Command) (bool, string, string, error) {
```
jakeyheath, line 157:
```suggestion
func VerifyHappyIsLockedVersion(cmd *cobra.Command) (string, string, error) {
```

I think having an error or not an error is good enough to verify. I think you can safely remove this bool

cli/cmd/version.go
jakeyheath: ahh here is the style guide I use : https://github.com/uber-go/guide/blob/master/style.md#reduce-nesting

cli/pkg/config/happy_version_lock.go
```diff
@@ -0,0 +1,92 @@
...
+
+	"github.com/pkg/errors"
+)
+
+type HappyVersionLockFile struct {
+	HappyVersion string
+	Path         string `json:"-"`
+}
+
+func NewHappyVersionLockFile(projectRoot string, requiredVersion string) (*HappyVersionLockFile, error) {
```
jakeyheath, line 18:
```suggestion
func NewHappyVersionLockFile(projectRoot, requiredVersion string) (*HappyVersionLockFile, error) {
```

cli/cmd/version.go
dspadea: In this case, the error would represent an error checking (likely reading the lock file). The bool says "Yes I checked, and yes it is the locked value (or it's not)".

cli/pkg/config/happy_version_lock.go
```diff
@@ -0,0 +1,92 @@
...
+
+	if projectRoot == "" {
+		return nil, errors.New("No projectRoot specified")
+	}
+
+	if requiredVersion == "" {
+		return nil, errors.New("No requiredVersion specified")
+	}
```
jakeyheath, line 26:
```suggestion
```

To avoid too much error checking all over the place, I usually only return errors when I really need to fail (like a network request has gone bad or something). For constructors like this where the parameters are required of the function, I think that is enough to the developer that they can't be empty. Remove the error from the return

cli/pkg/config/happy_version_lock.go
```diff
@@ -0,0 +1,92 @@
...
+
+	return &HappyVersionLockFile{
+		HappyVersion: requiredVersion,
+		Path:         path,
+	}, nil
+}
+
+func DoesHappyVersionLockFileExist(projectRoot string) bool {
+	filePath := calcHappyVersionPath(projectRoot)
+	_, err := os.Stat(filePath)
```
jakeyheath, line 38: I think os.Stat errors can be varied. You might need to do something like this:

~~~
_, err := os.Stat("/file/that/exists");

if err != nil {
	if os.IsNotExist(err) {
		// file does not exist, do something
	} else {
		// more serious errors
	}
}
~~~

cli/pkg/config/happy_version_lock.go
```diff
@@ -0,0 +1,92 @@
...
+	return err == nil
+}
+
+func LoadHappyVersionLockFile(projectRoot string) (*HappyVersionLockFile, error) {
+
+	filePath := calcHappyVersionPath(projectRoot)
+
+	versionFile, err := os.Open(filePath)
```
jakeyheath, line 46:
```suggestion
	versionFile, err := os.Open(calcHappyVersionPath(projectRoot))
```

cli/pkg/config/happy_version_lock.go
```diff
@@ -0,0 +1,92 @@
...
+	defer versionFile.Close()
+
+	contents, err := ioutil.ReadAll(versionFile)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to read happy version lock file")
+	}
+
+	hvlf := HappyVersionLockFile{}
+
+	err = json.Unmarshal(contents, &hvlf)
```
jakeyheath, line 59:
```suggestion
	hvlf := HappyVersionLockFile{}
	err = json.NewDecoder(versionFile).Decode(&hvlf)
```

when I already have a `reader` interface like a file, I opt for this method of decoding JSON because its all done in one step and I don't have to handle the the io.ReadAll errors. Its also a nice trick when you want to decode a stream of bytes like on a network request.

cli/cmd/version.go
dspadea: Done

cli/pkg/config/happy_version_lock.go
```diff
@@ -0,0 +1,92 @@
...
+		return errors.New(fmt.Sprintf("Could not create %s: %v", v.Path, err))
+	}
+
+	contents, err := json.MarshalIndent(&v, "", " ")
+	if err != nil {
+		return errors.Wrap(err, "could not marshal config file contents")
+	}
+
+	happyVersionFile.WriteString(string(contents))
+	happyVersionFile.Close()
```
jakeyheath, line 85: please make this a defer statement right after its created. Otherwise, if you hit one of the errors above you'll end up with a file that's not closed

cli/pkg/config/happy_version_lock.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	return &hvlf, nil
+}
+
+func (v *HappyVersionLockFile) Save() error {
+
+	if v.Path == "" {
+		return errors.New("Path is not set!")
+	}
```
jakeyheath, line 71:
```suggestion
```

similar to the error checking comment above, I think its safe to do your validation on the constructor and not have to check on every method

cli/pkg/config/happy_version_lock.go
```diff
@@ -0,0 +1,92 @@
...
+
+	happyVersionFile, err := os.Create(v.Path)
+
+	if err != nil {
+		return errors.New(fmt.Sprintf("Could not create %s: %v", v.Path, err))
```
jakeyheath, line 76:
```suggestion
		return errors.Wrapf("could not create %s: %v", v.Path, err)
```

cli/pkg/config/happy_version_lock.go
dspadea: Ok, removed

cli/cmd/version.go
```diff
@@ -67,8 +88,127 @@ func IsHappyOutdated(cmd *cobra.Command) (bool, *util.Release, *util.Release, er
 	cliVersion := util.GetVersion()
 	latestAvailableVersion, err := GetLatestAvailableVersion(cmd)
 	if err != nil {
-		return false, cliVersion, latestAvailableVersion, err
+		log.Debug("Error getting latest available version. Will fail silently.")
+		//return false, cliVersion, latestAvailableVersion, err
+		return false, cliVersion, cliVersion, nil // Lie.
```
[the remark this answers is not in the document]
dspadea: No - it's how it fails silently if it couldn't load a config to actually do a check. It just pretends it's up to date.

cli/pkg/config/happy_version_lock.go
dspadea: done

cli/pkg/config/happy_version_lock.go
dspadea: done

cli/pkg/config/happy_version_lock.go
dspadea: done

cli/pkg/config/happy_version_lock.go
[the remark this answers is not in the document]
dspadea: done

PR #1381, opened 2023-03-17
feat: configure WAF per happy-env-eks

kuannie1:


terraform/modules/happy-env-eks/outputs.tf
```diff
@@ -14,3 +14,12 @@ output "integration_secret" {
   value     = local.secret_string
   sensitive = true
 }
+
+output "panther_waf_configuration" {
```
alexlokshin-czi, line 18: Where is this going to be used?

alexlokshin-czi (approved):

terraform/modules/happy-env-eks/waf.tf
```diff
@@ -0,0 +1,5 @@
+module "regional-waf" {
+  count  = var.include_waf
```
alexlokshin-czi, line 2:
```suggest
  count  = var.include_waf ? 1 : 0
```

PR #1385, opened 2023-03-17
fix: front stack with WAF if it exists, ignore otherwise

kuannie1:
deleted a variable for simplicity. If a WAF config exists, add it! If not, ignore.

I'm confident about setting "waf_config" as the key because I've decided on that value [here](https://github.com/chanzuckerberg/happy/blob/main/terraform/modules/happy-env-eks/integration_secret.tf#L38-L43). We also set name & scope since those are required variables for the aws webacl data block [here](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/wafv2_web_acl)

Note: it works! I have the following screenshot to prove it. I created my own rdev stack in edu-ps-goal-setting-test to verify the WAF doesn't interfere with day-to-day.

Screenshot 2023-03-17 at 1 39 13 PM

This one shows that the WAF did get attached to my load balancers.

Screenshot 2023-03-17 at 1 39 55 PM

This one shows that I can navigate the site after logging in through okta. business can continue as usual

### DIFF
--- a/terraform/modules/happy-stack-eks/main.tf
+++ b/terraform/modules/happy-stack-eks/main.tf
@@ -133,7 +133,7 @@ locals {
 }
 
 data "aws_wafv2_web_acl" "happy_waf" {
-  count =  (local.waf_config != null) ? 1 : 0
+  count = (local.waf_config != null) ? 1 : 0
   name  = local.waf_config.name
   scope = local.waf_config.scope
 }


terraform/modules/happy-stack-eks/main.tf
```diff
@@ -183,7 +193,7 @@ module "services" {
 
   tags = local.secret["tags"]
 
-  regional_wafv2_arn = var.regional_wafv2_arn
+  regional_wafv2_arn = local.regional_waf_arn
```
alexlokshin-czi, line 196: Is null value accepted, if no WAF is configured?

alexlokshin-czi (approved):

PR #1388, opened 2023-03-17
feat: add happy_stacklist data item to provider

hspitzley-czi:


shared/model/app_stack.go
```diff
@@ -35,12 +35,12 @@ func NewAppStackFromAppStackPayload(payload AppStackPayload) *AppStack {
 	}
 }
 
-func MakeAppStackPayload(appName, env, stack, awsProfile, awsRegion, launghType, k8sNamespace, k8sClusterId string) AppStackPayload {
+func MakeAppStackPayload(appName, env, stack, awsProfile, awsRegion, launchType, k8sNamespace, k8sClusterId string) AppStackPayload {
```
alexlokshin-czi, line 38: Is it possible to rework this call to have less arguments? You can aggregate profile, region, namespace and clusterid...

alexlokshin-czi (approved):

shared/model/app_stack.go
hspitzley-czi: I didn't like how many args there are here either but I'm hesitant about the complexity that introduces (needing to create a data aggregate just to pass into a func that creates a data aggregate). Do you think it's worth it? Maybe it simplifies something else down the line? If so let's refactor in a follow-up since the change here is just fixing a typo

PR #1442, opened 2023-03-27
fix: tags flag for push and other create/update commands

jakeyheath:
This PR fixes a bug where the tags weren't properly applied when creating new stacks or pushing images. Here are some example commands that should work now:

~~~
happy push data-service --tags test1,test2
// will make a generated tag, tag named "test1", and tag named "test2"
~~~

~~~
happy create data-service --tag  test1
// will only make a tag called "test1" when pushing
~~~


cli/cmd/push.go
```diff
@@ -15,7 +15,7 @@ func init() {
 	config.ConfigureCmdWithBootstrapConfig(pushCmd)
 
 	pushCmd.Flags().StringVar(&sliceName, "slice", "", "The name of the slice you'd like to push to the registry.")
-	pushCmd.Flags().StringSliceVar(&tags, "tag", nil, "Extra tags to set for built images, comma-delimited (ex: tag1,tag2,tag3). We will, in addition, generate default tags automatically.")
+	pushCmd.Flags().StringSliceVar(&tags, "tags", nil, "Extra tags to set for built images, comma-delimited (ex: tag1,tag2,tag3). We will, in addition, generate default tags automatically.")
```
alexlokshin-czi, line 18: This can break some CICD invocations...

jakeyheath: ok, I have added an alias

alexlokshin-czi (approved):

PR #1447, opened 2023-03-28
fix: fix broken default tags in the default happy update/create cmds

jakeyheath:
I had messed up tags before, so that happy update was not using the right tags when deploying. This cleans up a lot of the tags functionality and works properly for happy update, create and push.


cli/cmd/update.go
```diff
@@ -134,9 +134,13 @@ func updateStackMeta(ctx context.Context, stackName string, happyClient *HappyCl
 		"happy/meta/configsecret": happyClient.HappyConfig.GetSecretId(),
 	})
 	if sliceDefaultTag != "" {
-		happyClient.Tag = sliceDefaultTag
+		happyClient.ArtifactBuilder.WithTags([]string{sliceDefaultTag})
```
alexlokshin-czi, line 137: Will this override the other tags?

cli/cmd/update.go
```diff
@@ -134,9 +134,13 @@ func updateStackMeta(ctx context.Context, stackName string, happyClient *HappyCl
...
 	}
-	err := stackMeta.Update(ctx, happyClient.Tag, happyClient.StackTags, "", happyClient.StackService)
+	// for updating and creating, there should only be one tag (either provided or generated)
+	if len(happyClient.ArtifactBuilder.GetTags()) != 1 {
+		return nil, errors.New("there should only be one tag when updating or creating a stack")
```
alexlokshin-czi, line 141: If we have a constraint like this, why do we allow a list of tags?

cli/cmd/update.go
jakeyheath: ArtifactBuilder is also used in the `happy push` command and that command allows you to specify multiple tags to push. So we need a way to specify multiple tags. In the update/create case though, you can only specify one tag, so that is why I check it here.

cli/cmd/update.go
jakeyheath: ArtifactBuilder is also used in the happy push command and that command allows you to specify multiple tags to push. So we need a way to specify multiple tags. In the update/create case though, you can only specify one tag, so that is why I check it here.

cli/cmd/update.go
jakeyheath: I am going to work with @dtsai-czi on cleaning up the slices work. I think they haven't been working for a bit, so I will investigate this again then.

alexlokshin-czi (approved):

PR #1469, opened 2023-04-04
feat: Expose stack TFE status, TFE Url, and Endpoints through HAPI

alexlokshin-czi:
Part of https://czi-tech.atlassian.net/browse/CCIE-924


api/pkg/api/app.go
```diff
@@ -85,25 +81,6 @@ func MakeApp(cfg *setup.Configuration) *APIApplication {
...
-		marshalled,
-		func(match []byte) []byte {
-			return bytes.ToLower(wordBarrierRegex.ReplaceAll(
-				match,
-				[]byte(`${1}_${2}`),
-			))
-		},
-	)
-	return converted, err
-}
```
hspitzley-czi, line 105: This was causing some unexpected behavior, putting the field in a struct tag works better

api/pkg/cmd/stack.go
```diff
@@ -3,15 +3,20 @@ package cmd
...
+	"github.com/chanzuckerberg/happy/shared/config"
 	"github.com/chanzuckerberg/happy/shared/model"
+	"github.com/chanzuckerberg/happy/shared/workspace_repo"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
 
 type StackManager interface {
-	GetAppStacks(context.Context, model.AppStackPayload) ([]*model.AppStack, error)
+	GetAppStacks(context.Context, model.AppStackPayload) ([]*model.AppStackResponse, error)
```
alexlokshin-czi, line 19: Do you think we need to start introducing shared models into the shared package?

hspitzley-czi: They are in shared package already actually

api/pkg/cmd/stack.go
```diff
@@ -44,18 +49,43 @@ func (s Stack) GetAppStacks(ctx context.Context, payload model.AppStackPayload)
...
+				stack.Endpoints = map[string]string{}
+				endpoints, err := workspace.GetEndpoints(ctx)
+				if err == nil {
+					stack.Endpoints = endpoints
+				}
+				stack.WorkspaceStatus = workspace.GetCurrentRunStatus(ctx)
+			}
+			stacks = append(stacks, stack)
+
+			wg.Done()
```
jakeyheath, line 85: make this a defer at the top of the function so we don't have to worry about changes to this code exiting early and not signaling the waiter should stop waiting

api/pkg/cmd/stack.go
```diff
@@ -44,18 +49,43 @@ func (s Stack) GetAppStacks(ctx context.Context, payload model.AppStackPayload)
...
 	for _, stackName := range stacklist {
-		appStack := model.MakeAppStack(payload.AppName, payload.Environment, stackName)
-		stacks = append(stacks, &appStack)
+		wg.Add(1)
+		go func(stackName string) {
+			stack := &model.AppStackResponse{
+				AppMetadata: *model.NewAppMetadata(payload.AppName, payload.Environment, stackName),
+			}
+			workspace, err := workspaceRepo.GetWorkspace(ctx, fmt.Sprintf("%s-%s", payload.AppMetadata.Environment, stackName))
+			if err == nil {
```
jakeyheath, line 74: rather than nesting, fail fast for error cases

```suggestion
			if err != nil {
			   return err
			}
```

jakeyheath: see also "golang.org/x/sync/errgroup". Might be helpful for capturing these errors a little better

api/pkg/cmd/stack.go
```diff
@@ -44,18 +49,43 @@ func (s Stack) GetAppStacks(ctx context.Context, payload model.AppStackPayload)
...
+				stack.WorkspaceUrl = workspace.GetWorkspaceUrl()
+				stack.Endpoints = map[string]string{}
+				endpoints, err := workspace.GetEndpoints(ctx)
+				if err == nil {
+					stack.Endpoints = endpoints
+				}
```
jakeyheath, line 80:
```suggestion
				stack.Endpoints, err := workspace.GetEndpoints(ctx)
				if err != nil {
					// TODO: don't forget to handle this error. 
					// if there is nothing to do, is it really an error?
				}
```

jakeyheath (approved):

api/pkg/cmd/stack.go
hspitzley-czi: good call, updated

api/pkg/cmd/stack.go
hspitzley-czi: This `if err == nil {` structure is intended in this case. It should just omit the workspace values when it can't find a workspace for a stack, similar to how the cli operates

I did make a small update to the behavior though, I added an error field on the response object that will be populated with the error when it's not nil

PR #1470, opened 2023-04-04
feat: add examples folder with first happy example project

jakeyheath:
## Overview

This PR adds an "examples" folder to the root of the happy project. All the sub directories will contain working happy examples that people can use to educate themselves on the various configurations and methods of using happy. This is the first step towards self-service and documentation, so we should always make sure these projects are working as expected and kept up-to-date.

All of these examples should be considered working. As in, if you are a CZI employee, you should be able to cd into the project and start running happy commands and they should all work. All the stacks are deployed into the czi-playground account. TODO: we need to add a clean up script that runs every night to clean up playground stacks.

## First Example

The first example is how to use a happy project as a container image base and have other happy projects consume that base image in their dockerfiles. This introduces a two new service_types: IMAGE_TEMPLATE and TARGET_GROUP_ONLY. IMAGE_TEMPLATE will be used when a happy project is only utilized for pushing base images. Not services will actually be deployed. Only an ECR is created. TARGET_GROUP_ONLY is a different feature that allows for users to selective choose if their service needs an ALB and target groups (the default) or just the target groups. This is handy if you have existing infra that you want to hook happy up to.


examples/image_pipeline/nodejs_base/.happy/terraform/envs/rdev/main.tf
```diff
@@ -0,0 +1,20 @@
+module "stack" {
+  source = "git@github.com:chanzuckerberg/happy//terraform/modules/happy-stack-eks?ref=heathj/examples-folder"
```
alexlokshin-czi, line 2: Can this be pointed to `main`?

alexlokshin-czi (approved):

PR #1488, opened 2023-04-06
feat: Exclude local terraform artifacts when zipping up the stack on create/update

alexlokshin-czi:


shared/util/local_processor.go
```diff
@@ -34,11 +35,19 @@ func (s *LocalProcessor) Tarzip(src string, f *os.File) error {
...
 		if err != nil {
 			return errors.Wrapf(err, "Unable to walk the file path %s", file)
 		}
 
 		if !fi.Mode().IsRegular() {
+			logrus.Debugf("Skipping file (%s) ...", fi.Name())
+			return nil
+		}
+
+		if fi.Name() == ".DS_Store" || fi.Name() == ".terraform" || fi.Name() == ".git" || fi.Name() == ".terraform.lock.hcl" || filepath.Ext(fi.Name()) == ".tar.gz" || dir == ".terraform" || dir == ".git" {
```
jakeyheath, line 49: can you add these into a slice and loop over them rather than a large conditional like this? something like

```suggestion
		ignored:=[]string{".DS_Store", "blah", "blah"}
		for _, i := range ignored {
		  return nil
		}
```

shared/util/local_processor.go
```diff
@@ -34,11 +35,19 @@ func (s *LocalProcessor) Tarzip(src string, f *os.File) error {
 	defer tw.Close()
 
 	return filepath.Walk(src, func(file string, fi os.FileInfo, err error) error {
```
jakeyheath, line 37: How will this work with directories? If there is a .terraform directory does it not go into that directory? I think if you want to skip a directory, you will need to do something like:
~~~golang
if info.IsDir() && info.Name() == subDirToSkip {
  fmt.Printf("skipping a dir without errors: %+v \n", info.Name())
  return filepath.SkipDir
}
~~~

alexlokshin-czi: Walk() is not recursive. The way it works is that it loops through the entire fileset across the entire folder tree, and skipping a dir entry will not skip the rest of the files in it. Instead I look at the entire file path and get the top level directory from it. If it is not allowed, I skip the entry. This way all files under `.terraform` will be skipped.

jakeyheath: looks like you fixed it up, but I think it is recursive: https://pkg.go.dev/path/filepath#WalkFunc

jakeyheath: the error type controls the flow of how the walker walks

alexlokshin-czi: Another revision.

jakeyheath (approved):

PR #1495, opened 2023-04-07
feat: Happy Service integrity check

alexlokshin-czi:


shared/util/tf_parser.go
```diff
@@ -0,0 +1,74 @@
...
+			f, diags := hclsyntax.ParseConfig(b, path, hcl.Pos{Line: 1, Column: 1})
+			if diags.HasErrors() {
+				return errors.Wrapf(diags.Errs()[0], "failed to parse %s", path)
+			}
+
+			content, _, contentDiags := f.Body.PartialContent(schema)
+			if contentDiags.HasErrors() {
+				return errors.New("Terraform code has errors")
+			}
+			for _, block := range content.Blocks {
```
jakeyheath, line 48: can you write this in a way that doesn't do the nesting like this? Essentially have the business logic in the same level of indenting, and if its an error continue the loop. Its harder to read like this

jakeyheath: aka. 

~~~golang
if block.Type != "module" {
  continue
}

if sourceAttr, ok := attrs["source"]; !ok {
 continue
}

// etc
//
// main business logic
~~~

cli/cmd/create.go
```diff
@@ -64,6 +64,7 @@ func runCreate(
 	ctx := cmd.Context()
 	message := workspace_repo.Message(fmt.Sprintf("Happy %s Create Stack [%s]", util.GetVersion().Version, stackName))
 	err = validate(
+		validateConfigurationIntegirty(ctx, happyClient),
```
jakeyheath, line 67: we should probably put this in the push command too

alexlokshin-czi: Done.

cli/cmd/init.go
```diff
@@ -180,6 +181,47 @@ func validateStackNameAvailable(ctx context.Context, stackService *stackservice.
...
+	return func() error {
+		// These services are configured in docker-compose.yml, and have their containers built
+		availableServices, err := happyClient.ArtifactBuilder.GetServices(ctx)
+		if err != nil {
+			return errors.Wrap(err, "unable to get available services")
+		}
+		// These services are configured in .happy/config.json
+		for _, service := range happyClient.HappyConfig.GetServices() {
+			if _, ok := availableServices[service]; !ok {
+				return errors.Errorf("service %s is configured in docker-compose.yml, but referenced in .happy/config.json", service)
```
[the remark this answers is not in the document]
alexlokshin-czi: You're right!

hspitzley-czi (approved):

PR #1526, opened 2023-04-10
feat: Add Happy User Documentation

dspadea:
This aims to add a first pass at Happy Path user documentation.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,0 +1,25 @@
+# What is Happy Path?
+
+Happy Path, affectionately known internally as "Happy", is an opinionated project lifecycle management tool which helps you
+with the complexities of managing software deployments across multiple environments and AWS accounts. It is designed to
+set you up early on the Happy Path to infrastructure success from the beginning of your project, rather than needing to
+figure it out as an afterthought when it's least convenient.
+
+# Table of Contents
+
+## Getting Started
+
+* [Install Happy Path](Installing_Happy_Path.md)
+* [Create a Happy Environment](Creating_a_Happy_Environment.md)
+
+## Using Happy Path
+
+* [Deploy An Application](Deploy_An_Application.md)
+
+## Using Kubernetes in a Happy Environment
+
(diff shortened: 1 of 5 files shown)


docs/Service_Discovery_In_EKS.md
```diff
@@ -0,0 +1,102 @@
+# Service Discovery in EKS
```
jakeyheath, line 1: All this information is very helpful. However, I wonder if we should focus on how people can leverage service discovery of their happy applications instead of the K8s details. I think a lot of our happy customers don't know anything about K8s (and they shouldn't have to). All they want to know if how they can talk to two services they deploy.

docs/README.md
```diff
@@ -0,0 +1,25 @@
...
+with the complexities of managing software deployments across multiple environments and AWS accounts. It is designed to
+set you up early on the Happy Path to infrastructure success from the beginning of your project, rather than needing to
+figure it out as an afterthought when it's least convenient.
+
+# Table of Contents
+
+## Getting Started
+
+* [Install Happy Path](Installing_Happy_Path.md)
+* [Create a Happy Environment](Creating_a_Happy_Environment.md)
```
jakeyheath, line 13: I can help fill out these sections in the next iteration

jakeyheath (approved):
NICE. DOCS! Lets merge this in and keep iterating on it.

docs/Service_Discovery_In_EKS.md
dspadea: Makes sense. I assumed that Core DNS would be central to that, but maybe I'm not sufficiently tying it into how to work with it practically. Happy to continue clarifying. These are more or less first passes.

docs/README.md
dspadea: Thank you! I'm a little stuck on some of the modules not being in the public repos. I've put the doc tickets into Blocked on the ticket to move the Happy TF modules. To Be Continued, I guess.

dspadea:
Additional docs here:

https://github.com/chanzuckerberg/happy/pull/1650

PR #1602, opened 2023-04-20
fix: remove skip; actually validate child modules

jakeyheath:
When a module accepts a configuration_alias, terraform isn't smart enough to know where the configuration_alias will come from. In the real world, it comes from the root module. However, when we want to `terraform validate` a child module, there is no root module. This results in the following errors mentioned here: https://github.com/chanzuckerberg/happy/pull/1600/files. The fix is actually to instantiate the module in a `.validate` folder and provide a configuration_alias (kind of like a mock). This `.validate` folder acts as the root module and will also validate all the child modules we want to actually validate.


terraform/modules/happy-env-ecs/.validate/main.tf
```diff
@@ -0,0 +1,27 @@
...
+    vpc_id                = "test"
+  }
+  tags = {
+    env       = "test"
+    managedBy = "teste"
+    owner     = "test"
+    project   = "test"
+    service   = "test"
+  }
+  providers = {
```
jakeyheath, line 20: This is how it is expecting the module be called, so you just have to pass in the alias so that validation on this module works properly. Think of this like a unit test. Please, no more skipping testing.

terraform/modules/happy-tfe-okta-service-account/.validate/main.tf
```diff
@@ -0,0 +1,18 @@
...
+    owner     = "test"
+    project   = "test"
+    service   = "test"
+  }
+  providers = {
+    aws.czi-si = aws.czi-si
+  }
+}
+
+provider "aws" {
```
jakeyheath, line 16: This is how it is expecting the module be called, so you just have to pass in the alias so that validation on this module works properly. Think of this like a unit test. Please, no more skipping testing.

alexlokshin-czi (approved):

PR #1607, opened 2023-04-20
fix: state drilling with dryrun; use context

jakeyheath:
This PR changes the way DryRun works. Instead of passing dryRun around the whole CLI, we store the value as part of the command's context. This will be used in the few places where dry run is needed and pulled out accordingly. I didn't change any business logic that utilizes dryRun, just the calling convention and how the value is attained.


cli/cmd/create.go
```diff
@@ -56,20 +57,20 @@ func runCreate(
...
 ) error {
 	stackName := args[0]
-	happyClient, err := makeHappyClient(cmd, sliceName, stackName, []string{tag}, createTag, dryRun, ModeCreate)
+	happyClient, err := makeHappyClient(cmd, sliceName, stackName, []string{tag}, createTag, ModeCreate)
 	if err != nil {
 		return errors.Wrap(err, "unable to initialize the happy client")
 	}
 
-	ctx := cmd.Context()
+	ctx := context.WithValue(cmd.Context(), options.DryRunKey, dryRun)
```
jakeyheath, line 65: dry run becomes a context value. most of this program doesn't need to know about it. For the pieces that do, pull it from the context

alexlokshin-czi: Last round of dryrun changes was causing stacks to be deleted after --dry-run was completing; do you see side effects?

cli/cmd/update.go
```diff
@@ -99,22 +99,11 @@ func updateStack(ctx context.Context, cmd *cobra.Command, stack *stackservice.St
 
 	// 2.) apply the terraform for the stack
 	stack = stack.WithMeta(stackMeta)
-	err = stack.Apply(ctx, makeWaitOptions(stack.Name, happyClient.HappyConfig, happyClient.AWSBackend), dryRun, workspace_repo.Message(fmt.Sprintf("Happy %s Update Stack [%s]", util.GetVersion().Version, stack.Name)))
+	err = stack.Apply(ctx, makeWaitOptions(stack.Name, happyClient.HappyConfig, happyClient.AWSBackend), workspace_repo.Message(fmt.Sprintf("Happy %s Update Stack [%s]", util.GetVersion().Version, stack.Name)))
 	if err != nil {
 		return errors.Wrap(err, "failed to apply the stack")
 	}
 
-	if dryRun {
```
alexlokshin-czi, line 107: This will cause the `happy create <STACK_NAME> --dry-run` to fail at the end with: 

```
unable to update the stack <STACK_NAME>: failed to run migrations: failed to get state for workspace
```

shared/workspace_repo/workspace.go
```diff
@@ -311,6 +311,10 @@ func (s *TFEWorkspace) WaitWithOptions(ctx context.Context, waitOptions options.
 		tfe.RunPlannedAndFinished: {},
 	}
 
+	dryRun, ok := ctx.Value(options.DryRunKey).(bool)
+	if !ok {
+		dryRun = false
+	}
 	if dryRun {
 		RunDoneStatuses = map[tfe.RunStatus]bool{
```
alexlokshin-czi, line 319: The first run that creates the workspace doesn't run in a dryrun mode. To address it, tfe.RunApplied should also be accepted as `RunDoneStatuses` here.

alexlokshin-czi: Also, TfeSuccessStatuses should include tfe.RunApplied for the same reason.

cli/cmd/create.go
alexlokshin-czi: This PR breaks `happy create <STACK_NAME> --dry-run`.

jakeyheath:
I think this should be fixed now. I was able to create a stack using both dry-run and non dry-run.

alexlokshin-czi (approved):

PR #1608, opened 2023-04-20
fix: needs_private_waf_attachment and needs_public_waf_attachment

Bento007:
- The variable was being set and used incorrectly.
- fix missing aws.czi-si provider in terraform/modules/happy-tfe-okta-service-account/ssm_writer.tf. The PR depends on https://github.com/chanzuckerberg/cztack/pull/478


terraform/modules/happy-tfe-okta-service-account/ssm_writer.tf
```diff
@@ -4,7 +4,7 @@ locals {
 }
 
 module "params" {
-  source  = "github.com/chanzuckerberg/cztack//aws-ssm-params-writer?ref=v0.43.1"
+  source  = "git@github.com:chanzuckerberg/cztack//aws-ssm-params-writer?ref=tsmith_aws_provider"
```
jakeyheath, line 7: 👍🏻 make sure to update this after the other one merges through. I'll approve for now

jakeyheath (approved):
TY!

terraform/modules/happy-tfe-okta-service-account/ssm_writer.tf
Bento007:
```suggestion
  source  = "git@github.com:chanzuckerberg/cztack//aws-ssm-params-writer?ref=v0.53.2"
```

PR #1611, opened 2023-04-21
chore: remove legacy unnecessary interfaces

jakeyheath:
Cleanup of exec code. CommandContext runs `LookPath`, so I have cleaned up all these duplicate references. Also, I have remove the LocalProfecessor interface since it doesn't have any implementation specific structs in it. Works just as well as a helper function and not clutter our structs.


alexlokshin-czi (approved):

alexlokshin-czi:
Mocks have to be re-generated.

PR #1617, opened 2023-04-24
chore: Always create iam service role

katyho:
### Summary
A service role isn't created unless the policy json is not empty. We'd like to be able to attach policies on the role outside of the `aws_iam_policy_json` field. It'd be helpful if this role was always created


jakeyheath (approved):
NICE!

PR #1648, opened 2023-04-25
fix: error for invalid policies on happy-iam-service-account-eks module

katyho:
### Summary
An empty string isn't a valid policy. This is causing the following error:
```
Error: "policy" contains an invalid JSON policy
```

### Test Plan
~pending testing on an existing happy example~

Created a stack on the `typical_app` example, and it works!
- an aws_iam_role resource is created
- no policy is created (so no policy errors occur during the TF plan or apply)


terraform/modules/happy-service-eks/variables.tf
```diff
@@ -114,7 +114,7 @@ variable "platform_architecture" {
 
 variable "aws_iam_policy_json" {
   type        = string
-  default     = ""
+  default     = "{}"
```
jakeyheath, line 117: I think this might also be invalid from my memory. Try something like the following:

https://stackoverflow.com/questions/62287436/is-it-possible-to-have-a-no-op-iam-policy

jgadling: Yep, as-is this errors out:
```
Error: updating IAM policy arn:aws:iam::xxxxx:policy/mysandbox-gql-dev-mysandbox: MalformedPolicyDocument: Syntax errors in policy.
```

katyho: Maybe instead we can always create the aws_iam_role, but create the aws_iam_role_policy and aws_iam_role_policy_attachment resources conditionally

jgadling: yeah, I have a branch that implements that here: https://github.com/chanzuckerberg/happy/compare/jgadling/targetfixes?expand=1

jakeyheath (approved):

PR #1653, opened 2023-04-25
fix: Trigger a happy-env-eks release

alexlokshin-czi:


jakeyheath (changes requested):
needs to be a fix or feat otherwise the version won't bump

PR #1658, opened 2023-04-25
chore: improve validation for target_group_only paths.

jgadling:
TARGET_GROUP_ONLY services do not need their healthcheck prefixes to be a subpath of their service path. This PR clarifies that.


jakeyheath (dismissed):

terraform/modules/happy-stack-eks/variables.tf
```diff
@@ -80,7 +80,8 @@ variable "services" {
     error_message = "The service_type 'TARGET_GROUP_ONLY' requires an alb"
   }
   validation {
-    condition     = alltrue([for k, v in var.services : startswith(v.health_check_path, trimsuffix(v.path, "*"))])
+    # The health check prefix needs to contain the service path for CONTEXT services, but not TARGET_GROUP_ONLY services.
+    condition     = alltrue([for k, v in var.services : anytrue([startswith(v.health_check_path, trimsuffix(v.path, "*")), v.service_type != "CONTEXTUAL"])])
```
jakeyheath, line 84: did you mean to say `v.service_type != "TARGET_GROUP_ONLY"`? I don't know what contextual is

jgadling: Oops, good catch -- I updated it to be CONTEXT.
This validation is unnecessary for TARGET_GROUP_ONLY services. I'm not sure if it's needed for CONTEXT services, but the validator says "health check path must start with the service path unless the service type is something other than CONTEXT"

jakeyheath (approved):

PR #1659, opened 2023-04-25
feat: CCIE-960 do not require docker to be running for functions that don t use it

dspadea:
Removed preflight checks for list, version, and help. 

List still requires AWS and Terraform to be installed, but does not require Docker to be installed or running.

I'd like an opinion on the list of checks for each command, but the mechanism is there now to easily change the checklist. We can iterate on it pretty easily.


shared/util/preconditions.go
```diff
@@ -12,63 +12,103 @@ import (
...
+		valid, reasons := dockerComposeMinVersion.Validate(dockerComposeVersion)
+		if !valid {
+			errs = multierror.Append(
+				errs,
+				errors.Errorf("docker compose >= V2 required but %s was detected, please follow https://docs.docker.com/compose/cli-command/", version),
+			)
+			for _, reason := range reasons {
+				errs = multierror.Append(errs, reason)
+			}
+		}
```
jakeyheath, line 77: I would recommend breaking up these as individual functions and using them in the respective commands validation checks. See link above.

jakeyheath:
I love this feature, thank you for writing. A suggest a minor refactor where we put the validation on the specific command so that the code is easier to follow and manage on a command by command basis.

cli/cmd/root.go
```diff
@@ -119,3 +119,29 @@ func Execute() error {
 
 	return nil
 }
+
+func GetPreflightChecks(commandName string) *util.ValidationCheckList {
+	// This is sort of a weird way to do it, up front for all commands.
+	// Really, each command should have its own list of checks built into its
+	// handler. I'm doing it this way to keep with the way the program
+	// is currently structured. It may be worth refactoring this in the future.
+
```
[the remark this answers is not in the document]
dspadea: Thanks Jake. I'll go ahead and refactor this. I was trying to have a light touch, but in this case, I do think it's going to get a bit messy over time.

shared/util/preconditions.go
dspadea: Agreed. Ok - I'll work on it. Should be done with it today.

dspadea:
@jakeyheath This more what you were thinking?

cli/cmd/config.go
```diff
@@ -75,11 +75,27 @@ func ValidateConfigFeature(cmd *cobra.Command, args []string) error {
...
-	SilenceUsage:      true,
-	PersistentPreRunE: ValidateConfigFeature,
+	Use:          "config",
+	Short:        "modify app configs",
+	Long:         "Create, Read, Update, and Delete app configs for environment '{env}'",
+	SilenceUsage: true,
+	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+
+		if err := ValidateConfigFeature(cmd, args); err != nil {
+			return ValidateConfigFeature(cmd, args)
```
jakeyheath, line 85:
```suggestion
			return err
```

shared/util/preconditions.go
```diff
@@ -12,17 +12,61 @@ import (
...
+	DockerInstalled                  ValidationCallback
+	DockerEngineRunning              ValidationCallback
+	AwsInstalled                     ValidationCallback
+	TerraformInstalled               ValidationCallback
+	AwsSessionManagerPluginInstalled ValidationCallback
+}
+
+// Takes a list of callbacks to run. Canned ones are in the ValidationCheckList struct,
+// but nothing stops you from adding custom ones from other packages.
+func ValidateEnvironment(ctx context.Context, validations []ValidationCallback) error {
```
jakeyheath, line 28:
```suggestion
func ValidateEnvironment(ctx context.Context, validations ...ValidationCallback) error {
```

cli/cmd/build.go
```diff
@@ -19,6 +20,18 @@ var buildCmd = &cobra.Command{
...
+		checklist := util.NewValidationCheckList()
+		return util.ValidateEnvironment(cmd.Context(),
+			[]util.ValidationCallback{
+				checklist.DockerEngineRunning,
+				checklist.MinDockerComposeVersion,
+				checklist.DockerInstalled,
+				checklist.TerraformInstalled,
+				checklist.AwsInstalled,
+				checklist.AwsSessionManagerPluginInstalled,
+			})
```
jakeyheath, line 33: if you change the signature of this method to be varadic, you can do something like this:

```suggestion
				checklist.DockerEngineRunning,
				checklist.MinDockerComposeVersion,
				checklist.DockerInstalled,
				checklist.TerraformInstalled,
				checklist.AwsInstalled,
				checklist.AwsSessionManagerPluginInstalled,
			)
```

which I think reads a little easier

shared/util/preconditions.go
```diff
@@ -12,17 +12,61 @@ import (
...
+type ValidationCallback = func(context.Context) error
+
+type ValidationCheckList struct {
+	MinDockerComposeVersion          ValidationCallback
+	DockerInstalled                  ValidationCallback
+	DockerEngineRunning              ValidationCallback
+	AwsInstalled                     ValidationCallback
+	TerraformInstalled               ValidationCallback
+	AwsSessionManagerPluginInstalled ValidationCallback
+}
```
jakeyheath, line 24: I feel this struct is a bit unnecessary, but I have very weak feelings about it

alexlokshin-czi (approved):

shared/util/preconditions.go
alexlokshin-czi: Can this be a map?

shared/util/preconditions.go
```diff
@@ -45,21 +89,42 @@ func ValidateEnvironment(ctx context.Context) error {
...
+	_, err := exec.LookPath("aws")
 	if err != nil {
 		errs = multierror.Append(errs, errors.Wrap(err, "could not find aws cli in path, run 'brew install awscli' or follow https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html"))
 	}
 
-	_, err = exec.LookPath("terraform")
+	return errs.ErrorOrNil()
+}
+
+func ValidateAwsSessionManagerPluginInstalled(ctx context.Context) error {
```
alexlokshin-czi, line 105: This is an ECS specific tool, we probably need to conditionally validate.

shared/util/preconditions.go
dspadea: It can be anything, or nothing. The idea was to "bundle" them up to make code completion work better. Otherwise it's just a bunch of functions in the util namespace. There's nothing special about it otherwise.

shared/util/preconditions.go
dspadea: I like that idea. Hadn't thought of it.

shared/util/preconditions.go
dspadea: I'd say for this pass, just always validate wherever it might be used by a given command. Trying to figure out when it might be required seems potentially fraught. If it really annoys people, we can add that logic later.

cli/cmd/config.go
dspadea: changed

PR #1662, opened 2023-04-26
feat: Happy CLI integration test

alexlokshin-czi:
This PR will create an integration test that runs through typical happy user scenarios, like `happy create`, or `happy logs`, wrapping up by deleting the `integration-test` stack.


cli/pkg/stack_mgr/stack.go
```diff
@@ -296,16 +296,6 @@ func (s *Stack) Apply(ctx context.Context, waitOptions options.WaitOptions, runO
 
 	logrus.Debugf("will use tf bundle found at %s", srcDir)
 
-	tempFile, err := os.CreateTemp("", "happy_tfe.*.tar.gz")
```
alexlokshin-czi, line 299: This block of code was producing a gz file that wasn't used anywhere.

.github/workflows/integration-test.yml
```diff
@@ -0,0 +1,91 @@
+on:
+    push:
+      branches:
+        - main
+        - alokshin/integration-test
+name: integration-test
+concurrency:
```
alexlokshin-czi, line 7: To prevent other workflow runs to affect the integration-test stack.

examples/integration_test/docker-compose.yml
```diff
@@ -0,0 +1,11 @@
+version: "3.8"
+
+services:
+  frontend:
+    image: "frontend"
+    profiles: [ "*" ]
+    platform: linux/amd64
```
alexlokshin-czi, line 7: This Github runner can only create amd64 images, because qemu is not configured.

cli/pkg/stack_mgr/stack.go
jakeyheath: I've seen this! I always assumed I didn't know what was going on

jakeyheath (approved):
Looks great!

PR #1663, opened 2023-04-26
fix: workspaces with no state will not have info

jakeyheath:
This PR fixes an issue in the API where if the TFE workspace doesn't have any state (as in, there have been no applies or if there was an error on their first apply), the state information is not populated. This leads to a nil pointer exception. This PR handles this by always checking for state and returning empty information if no state.

### DIFF
--- a/api/pkg/cmd/stack.go
+++ b/api/pkg/cmd/stack.go
@@ -62,36 +62,40 @@ func enrichStacklistMetadata(ctx context.Context, stacklist []string, payload mo
 	).WithTFEToken(setup.GetConfiguration().TFE.Token)
 	wg := sync.WaitGroup{}
 
-	stacks := []*model.AppStackResponse{}
-
-	for _, stackName := range stacklist {
+	stackInfos := make([]*model.AppStackResponse, len(stacklist))
+	for i, stackName := range stacklist {
 		wg.Add(1)
-		go func(stackName string) {
+		go func(id int, stackName string) {
 			defer wg.Done()
 
-			stack := &model.AppStackResponse{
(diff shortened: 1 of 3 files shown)


api/pkg/cmd/stack.go
```diff
@@ -62,36 +62,40 @@ func enrichStacklistMetadata(ctx context.Context, stacklist []string, payload mo
...
 		wg.Add(1)
-		go func(stackName string) {
+		go func(id int, stackName string) {
 			defer wg.Done()
 
-			stack := &model.AppStackResponse{
-				AppMetadata: *model.NewAppMetadata(payload.AppName, payload.Environment, stackName),
-			}
-			workspace, err := workspaceRepo.GetWorkspace(ctx, fmt.Sprintf("%s-%s", payload.AppMetadata.Environment, stackName))
+			stackInfo, err := getStackInfo(ctx, payload, stackName, workspaceRepo)
```
jakeyheath, line 71: small goroutine refactor. I try to have as little lines in these goroutine functions as possible. it might seem silly but helps keep the concurrency pieces easy to read. Having a function that can return an error also makes the conditional logic simpler.

shared/workspace_repo/workspace.go
```diff
@@ -595,6 +600,11 @@ func (s *TFEWorkspace) GetCurrentRunStatus(ctx context.Context) string {
 }
 
 func (s *TFEWorkspace) GetCurrentRunUrl(ctx context.Context) string {
+	state, err := s.HasState(ctx)
+	if err != nil || !state {
+		return s.GetWorkspaceUrl()
```
jakeyheath, line 605: this kind of hides an error but I don't think we need to introduce another error.

dtsai-czi (approved):

PR #1683, opened 2023-05-02
fix: thread CI role through happy env

jakeyheath:
This PR refactors how the Github Action CI role is created and provisioned for happy projects. For EKS happy, the CI role needs to be created at the same time as the cluster so that the role can be added to the permission model of the cluster. Therefore, we need to be able to pass the role through as an argument and attach policies to it rather that create it in the happy modules. This allows us to keep a central point of truth for which roles have access to the cluster.

## Minor other things

* Moved ECR module to cztack and removed template dependency
* Split the ECS and EKS components of the CI role so we don't have to have unnecessary fields for the role

## References

* https://github.com/chanzuckerberg/shared-infra/pull/7746


terraform/modules/happy-env-eks/integration_secret.tf
```diff
@@ -11,8 +11,8 @@ locals {
     eks_cluster     = var.eks-cluster
     tags            = var.tags
     certificate_arn = module.cert.arn
-
-    ecrs = { for name, ecr in module.ecrs : name => { "url" : ecr.repository_url } }
+    ci_roles        = var.github_actions_roles
```
jakeyheath, line 14: TODO: make CLI function to read this so it easy for developers to get their CI role ARN

terraform/modules/happy-github-ci-role-eks/main.tf
```diff
@@ -12,47 +12,12 @@ data "aws_iam_policy_document" "eks" {
...
   }
 }
 
 resource "aws_iam_role_policy" "eks" {
   name   = "gh_actions_eks_describe_cluster_${random_pet.this.id}"
   policy = data.aws_iam_policy_document.eks.json
   role   = random_pet.this.keepers.role_name
-}
-
-data "kubernetes_config_map" "aws-auth" {
```
jakeyheath, line 25: this did not work, removing

alexlokshin-czi (approved):

terraform/modules/happy-env-eks/integration_secret.tf
jakeyheath: https://github.com/chanzuckerberg/happy/pull/1686

PR #1684, opened 2023-05-02
feat: Implement Happy cli basic mode (terraform code is generated)

alexlokshin-czi:
This new feature is invoked by `happy infra generate` and produces an entire set up HCL files for a given environment.


cli/pkg/stack_mgr/stack_service.go
```diff
@@ -417,3 +421,78 @@ func (s *StackService) getDistributedLock() (*backend.DistributedLock, error) {
 	lockConfig := backend.DistributedLockConfig{DynamodbTableName: s.backend.Conf().GetDynamoLocktableName()}
 	return backend.NewDistributedLock(&lockConfig, s.backend.GetDynamoDBClient())
 }
+
+func (s *StackService) Generate(ctx context.Context) error {
+	moduleSource := "git@github.com:chanzuckerberg/happy//terraform/modules/happy-stack-%s?ref=main"
```
alexlokshin-czi, line 426: TODO: Make this a configuration parameter.

shared/util/tf/tf_generator.go
```diff
@@ -0,0 +1,406 @@
...
+		Version: ">= 3.20.0",
+	},
+	{
+		Name:    "happy",
+		Source:  "chanzuckerberg/happy",
+		Version: ">= 0.53.5",
+	},
+}
+
+var requiredVariables []variable = []variable{
```
alexlokshin-czi, line 55: These have to be made configurable, but where to get the configuration from?

shared/util/tf/tf_generator.go
```diff
@@ -0,0 +1,406 @@
...
+}
+
+type variable struct {
+	Name        string
+	Type        string
+	Description string
+	Default     cty.Value
+}
+
+var requiredProviders []provider = []provider{
```
alexlokshin-czi, line 32: How to maintain this configuration?

shared/util/tf/tf_generator.go
```diff
@@ -0,0 +1,439 @@
...
+	defer tfFile.Close()
+	hclFile := hclwrite.NewEmptyFile()
+
+	rootBody := hclFile.Body()
+	err = tf.generateAwsProvider(rootBody, "", "${var.aws_account_id}", "${var.aws_role}")
+	if err != nil {
+		return errors.Wrap(err, "Unable to generate HCL code for AWS provider")
+	}
+
+	err = tf.generateAwsProvider(rootBody, "czi-si", "626314663667", "tfe-si")
```
alexlokshin-czi, line 268: TODO: Pull the account id out.

cli/cmd/infra_generate.go
```diff
@@ -0,0 +1,41 @@
...
+
+func init() {
+	infraCmd.AddCommand(infraGenerateCmd)
+	config.ConfigureCmdWithBootstrapConfig(infraGenerateCmd)
+}
+
+var infraGenerateCmd = &cobra.Command{
+	Use:          "generate",
+	Short:        "Generate HCL code",
+	Long:         "Generate HCL code in environment '{env}'",
```
jakeyheath, line 19:
```suggestion
	Short:        "Generate Happy Stack HCL code",
	Long:         "Generate Happy Stack HCL code in environment '{env}'",
```

Make it more clear what this HCL code is

cli/cmd/infra_generate.go
```diff
@@ -0,0 +1,41 @@
...
+		ctx := cmd.Context()
+
+		happyConfig, err := config.GetHappyConfigForCmd(cmd)
+		if err != nil {
+			return err
+		}
+
+		stackService := stackservice.NewStackService().WithHappyConfig(happyConfig)
+
+		logrus.Info("Generating HCL code")
```
jakeyheath, line 38:
```suggestion
		logrus.Debug("Generating HCL code")
```

cli/pkg/stack_mgr/stack_service.go
```diff
@@ -417,3 +422,95 @@ func (s *StackService) getDistributedLock() (*backend.DistributedLock, error) {
...
+	} else {
+		moduleSource = fmt.Sprintf(moduleSource, "ecs")
+	}
+	if stackConfig.Source != nil && len(*stackConfig.Source) > 0 {
+		moduleSource = *stackConfig.Source
+	}
+
+	_, modulePath, _, err := tf.ParseModuleSource(moduleSource)
+	if err != nil {
+		return errors.Wrap(err, "Unable to parse module path out")
```
jakeyheath, line 443: nit: wrapped errors should start with an uppercase

jakeyheath:
```suggestion
		return errors.Wrap(err, "unable to parse module path out")
```

jakeyheath: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

jakeyheath: make this same edit to the rest of the errors

cli/pkg/stack_mgr/stack_service.go
```diff
@@ -417,3 +422,95 @@ func (s *StackService) getDistributedLock() (*backend.DistributedLock, error) {
...
+	gen := tf.NewTfGenerator(s.GetConfig())
+
+	if _, err := os.Stat(srcDir); os.IsNotExist(err) {
+		err = os.MkdirAll(srcDir, 0777)
+		if err != nil {
+			return errors.Wrapf(err, "Unable to create terraform directory: %s", srcDir)
+		}
+	}
+
+	logrus.Infof("Generating terraform files in %s", srcDir)
```
jakeyheath, line 488:
```suggestion
	logrus.Debugf("Generating terraform files in %s", srcDir)
```

shared/util/cloner.go
```diff
@@ -0,0 +1,38 @@
...
+		return err
+	}
+
+	if err = json.Unmarshal(data, &dst); err != nil {
+		return err
+	}
+	return nil
+}
+
+func DeepMerge(dst, src map[string]interface{}) map[string]interface{} {
```
jakeyheath, line 20: why have `dst` and also a return value?

shared/util/cloner.go
```diff
@@ -0,0 +1,38 @@
...
+	data, err := json.Marshal(src)
+	if err != nil {
+		return err
+	}
+
+	if err = json.Unmarshal(data, &dst); err != nil {
+		return err
+	}
+	return nil
+}
```
jakeyheath, line 18: This might be a good use case for golang generics rather than `interface{}`. They've also updated `interface{}` to be called `any`

shared/util/tf/testdata/outputs.tf
```diff
@@ -0,0 +1,18 @@
+output "dashboard" {
+  sensitive = false
+  value     = module.stack.dashboard
+}
+output "service_ecrs" {
+  sensitive = false
+  value     = module.stack.service_ecrs
+}
+output "service_endpoints" {
```
jakeyheath, line 9:
```suggestion
output "service_urls" {
```

jakeyheath: We use service_urls, not service_endpoints in the Happy CLI code

jakeyheath: https://github.com/chanzuckerberg/happy/blob/45b27099d6504d16f58789e81bca7d7ef1c7e2b0/shared/workspace_repo/workspace.go#L667

shared/util/tf/testdata/variables.tf
```diff
@@ -0,0 +1,34 @@
...
+}
+variable "stack_name" {
+  type        = string
+  description = "Happy Path stack name"
+}
+variable "wait_for_steady_state" {
+  type        = bool
+  description = "Should terraform block until k8s deployment reaches a steady state?"
+  default     = true
+}
```
jakeyheath, line 34: this is not used anywhere I don't think

shared/util/tf/tf_parser.go
```diff
@@ -0,0 +1,348 @@
...
+			Type:       "output",
+			LabelNames: []string{"name"},
+		},
+	},
+}
+
+func (tf TfParser) ParseServices(dir string) (map[string]bool, error) {
+	var services map[string]bool = map[string]bool{}
+
+	err := filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
```
jakeyheath, line 70: we should use the newer function https://pkg.go.dev/path/filepath#WalkDir

jakeyheath: `Walk is less efficient than WalkDir, introduced in Go 1.16, which avoids calling os.Lstat on every visited file or directory.`

shared/util/tf/tf_parser.go
```diff
@@ -0,0 +1,348 @@
...
+		if err != nil {
+			return err
+		}
+		if info.IsDir() {
+			if info.Name() == ".terraform" || info.Name() == ".git" {
+				return filepath.SkipDir
+			}
+		}
+		if !info.IsDir() && filepath.Ext(path) == ".tf" {
```
jakeyheath, line 79:
```suggestion
		if info.IsDir() || filepath.Ext(path) != ".tf" { return nil } 
```

then the rest of the logic. Avoid nesting when possible

https://github.com/uber-go/guide/blob/master/style.md#reduce-nesting

jakeyheath:
Can you add a bit more of a description on how to use this feature, how to make edits to the template going forward, etc? Just a bit more information please.

shared/util/tf/tf_parser.go
```diff
@@ -0,0 +1,348 @@
...
+				LabelNames: []string{"name"},
+			},
+			{
+				Type:       "validation",
+				LabelNames: []string{"name"},
+			},
+		},
+	}
+
+	err := filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
```
jakeyheath, line 154:
```suggestion
	err := filepath.WalkDir(dir, func(path string, info os.FileInfo, err error) error {
```

shared/util/tf/tf_parser.go
```diff
@@ -0,0 +1,348 @@
...
+		if err != nil {
+			return err
+		}
+		if info.IsDir() {
+			if info.Name() == ".terraform" || info.Name() == ".git" {
+				return filepath.SkipDir
+			}
+		}
+		if !info.IsDir() && filepath.Ext(path) == ".tf" {
```
jakeyheath, line 163: nesting comment from above

shared/util/tf/tf_parser.go
```diff
@@ -0,0 +1,348 @@
...
+				return errors.Wrapf(diags.Errs()[0], "failed to parse %s", path)
+			}
+
+			content, _, diags := f.Body.PartialContent(outputBlockSchema)
+			if diags.HasErrors() {
+				return errors.New("Terraform code has errors")
+			}
+			for _, block := range content.Blocks {
+				if block.Type != "output" {
+					continue
```
jakeyheath, line 289: I feel like there is a lot repeated code between these methods

shared/util/cloner.go
alexlokshin-czi: Removed.

shared/util/tf/tf_generator.go
```diff
@@ -0,0 +1,439 @@
...
+		Name:        "stack_name",
+		Type:        "string",
+		Description: "Happy Path stack name",
+	},
+	{
+		Name:        "wait_for_steady_state",
+		Type:        "bool",
+		Description: "Should terraform block until k8s deployment reaches a steady state?",
+		Default:     cty.BoolVal(true),
+	},
```
jakeyheath, line 98:
```suggestion
	
```

shared/util/cloner.go
alexlokshin-czi: Since we don't really manipulate any functions generics might expose, I'll change the parameter type to `any`.

shared/util/tf/tf_generator.go
```diff
@@ -0,0 +1,439 @@
...
+	_, stackConfig, err := tf.happyConfig.GetStackConfig()
+	if err != nil {
+		return errors.Wrap(err, "Unable to get stack config")
+	}
+
+	tfFile, err := os.Create(filepath.Join(srcDir, "main.tf"))
+	if err != nil {
+		return errors.Wrap(err, "Unable to generate HCL code")
+	}
+	defer tfFile.Close()
```
jakeyheath, line 123: we should probably prepend

`# Auto-generated by "happy infra init". Use happy infra generate or happy infra init to regenerate`

or something along those line so people know the file was generated

shared/util/tf/testdata/outputs.tf
alexlokshin-czi: We do, but the module doesn't have an output of `service_urls`. https://github.com/chanzuckerberg/happy/blob/45b27099d6504d16f58789e81bca7d7ef1c7e2b0/terraform/modules/happy-stack-eks/outputs.tf#LL1C1-L1C1

shared/util/tf/tf_generator.go
```diff
@@ -0,0 +1,439 @@
...
+	happyConfig *config.HappyConfig
+}
+
+func NewTfGenerator(happyConfig *config.HappyConfig) TfGenerator {
+	return TfGenerator{
+		happyConfig: happyConfig,
+	}
+}
+
+func (tf *TfGenerator) GenerateMain(srcDir, moduleSource string, variables []Variable) error {
```
jakeyheath, line 113: I'll be honest, this function is not my favorite. I would have no idea how to make changes to it or update it and I don't necessarily see the advantages of use an HCL generator over a straight go template. We have this very large switch case anyway, so even though we are doing some dynamic stuff, we'll still need to maintain this switch statement, which is no different then maintaining a template.

You're going to have to convince me on this approach.

shared/util/tf/testdata/outputs.tf
alexlokshin-czi: Best I can do is add support for both `service_urls` (manually created HCL) and `service_endpoints` (automatically generated) outputs.

shared/util/tf/tf_generator.go
jakeyheath: Can you go over what steps I need to do if I want to upgrade my TF stack version?

And also, what steps would I need to do as a happy developer if we introduced a new stack variable or removed a stack variable? Sounds like we'd need to make changes to the StackConfig struct right?

jakeyheath: For instance, why not do what we do with fogg apply for module_source, and then add a JSON file with this switch case? I feel like editing a JSON file would be much simpler

shared/util/tf/testdata/variables.tf
alexlokshin-czi: This is just test data.

shared/util/tf/tf_parser.go
alexlokshin-czi: Changed.

shared/util/tf/tf_generator.go
```diff
@@ -0,0 +1,439 @@
...
+func (tf TfGenerator) generateAwsProvider(rootBody *hclwrite.Body, alias, accountIdExpr, roleExpr string) error {
+	awsProviderBody := rootBody.AppendNewBlock("provider", []string{"aws"}).Body()
+	if alias != "" {
+		awsProviderBody.SetAttributeValue("alias", cty.StringVal(alias))
+	}
+	awsProviderBody.SetAttributeValue("region", cty.StringVal(*tf.happyConfig.AwsRegion()))
+
+	assumeRoleBlockBody := awsProviderBody.AppendNewBlock("assume_role", nil).Body()
+	assumeRoleBlockBody.SetAttributeRaw("role_arn", tokens(fmt.Sprintf("\"arn:aws:iam::%s:role/%s\"", accountIdExpr, roleExpr)))
+	awsProviderBody.SetAttributeRaw("allowed_account_ids", tokens(fmt.Sprintf("[\"%s\"]", accountIdExpr)))
```
jakeyheath, line 312: right now the provider.tf is a static file. I think we should use a template for it since there is no dynamic configuration.

shared/util/tf/tf_generator.go
```diff
@@ -0,0 +1,439 @@
...
+	hclFile := hclwrite.NewEmptyFile()
+
+	rootBody := hclFile.Body()
+	err = tf.generateAwsProvider(rootBody, "", "${var.aws_account_id}", "${var.aws_role}")
+	if err != nil {
+		return errors.Wrap(err, "Unable to generate HCL code for AWS provider")
+	}
+
+	err = tf.generateAwsProvider(rootBody, "czi-si", "626314663667", "tfe-si")
+	if err != nil {
```
jakeyheath, line 269: right now the provider.tf is a static file. I think we should use a template for it since there is no dynamic configuration.

shared/util/tf/tf_generator.go
```diff
@@ -0,0 +1,439 @@
...
+			"source":  cty.StringVal(provider.Source),
+			"version": cty.StringVal(provider.Version),
+		})
+		requiredProvidersBody.SetAttributeValue(provider.Name, p)
+
+	}
+
+	_, err = tfFile.Write(hclFile.Bytes())
+
+	return err
```
jakeyheath, line 340: this is a static file, I don't think we need to use an HCL generator for this

shared/util/tf/tf_generator.go
```diff
@@ -0,0 +1,439 @@
...
+			hcl.TraverseRoot{Name: "module"},
+			hcl.TraverseAttr{Name: "stack"},
+			hcl.TraverseAttr{Name: output.Name},
+		})
+		moduleOutputBody.SetAttributeRaw("value", tokens)
+	}
+
+	_, err = tfFile.Write(hclFile.Bytes())
+
+	return err
```
jakeyheath, line 373: static file, same comment as above

shared/util/tf/tf_generator.go
```diff
@@ -0,0 +1,439 @@
...
+		variableBody.SetAttributeRaw("type", tokens)
+		variableBody.SetAttributeValue("description", cty.StringVal(variable.Description))
+		if !variable.Default.IsNull() {
+			variableBody.SetAttributeValue("default", variable.Default)
+		}
+	}
+
+	_, err = tfFile.Write(hclFile.Bytes())
+
+	return err
```
jakeyheath, line 399: static file

shared/util/tf/tf_parser.go
alexlokshin-czi: Changed.

shared/util/tf/tf_parser.go
alexlokshin-czi: I will save refactoring for a separate PR. Flow here seems similar but it isn't. I most likely will combine all of the ParseX() methods into one, but this requires a lot of testing.

shared/util/tf/tf_generator.go
alexlokshin-czi: This variable is present in `happy_env_ecs` module, I don't think we can drop it.

shared/util/tf/tf_generator.go
alexlokshin-czi: I can collapse the switch case further, as many of the variables are handled the same way, but not all. There are unique cases:

* References to variables
* Variables with values that come from happy config or module defaults
* Variables with static values
* Special cases (`services`) which are nested of unknown structure

The main goal is to have a source of truth, which template is not. Let's go over it tomorrow.

shared/util/tf/tf_generator.go
alexlokshin-czi: There is. Region is dynamic.

shared/util/tf/tf_generator.go
alexlokshin-czi: Region is dynamic.

shared/util/tf/tf_generator.go
alexlokshin-czi: I need to mull it over.

shared/util/tf/tf_generator.go
alexlokshin-czi: This is not static. It pull module outputs and exposes them. If the module changes, outputs will change as well. For example, our current stack level outputs.tf is out of sync with the happy-env-* module outputs.

shared/util/tf/tf_generator.go
alexlokshin-czi: Will tackle at a later point.

shared/util/tf/tf_generator.go
alexlokshin-czi: Upon further thought, I'd like to keep it generated. This PR is only the first of the series, this will continue to evolve.

shared/config/happy_config.go
```diff
@@ -29,6 +29,7 @@ type Environment struct {
 	AutoRunMigrations  bool            `yaml:"auto_run_migrations"`
 	TaskLaunchType     util.LaunchType `yaml:"task_launch_type"`
 	LogGroupPrefix     string          `yaml:"log_group_prefix"`
+	StackOverrides     map[string]any  `yaml:"stack_overrides"`
```
alexlokshin-czi, line 32: Any config values can be stored here to override the StackDefaults; and both are used to override module defaults.

shared/util/cloner.go
```diff
@@ -0,0 +1,56 @@
...
+			if _, ok := dst[k]; ok {
+				if v != nil {
+					if o, ok := dst[k]; ok {
+						t1 := reflect.TypeOf(o)
+						if t1 != t {
+							dstTypeName := "unknown"
+							if t1 != nil {
+								dstTypeName = t1.Name()
+							}
+							return errors.Errorf("Destination type (%s) is not the same as source type (%s) for field '%s'.", dstTypeName, t.Name(), k)
```
alexlokshin-czi, line 47: This is used to enforce type consistency between source and destination.

shared/util/tf/tf_parser.go
```diff
@@ -0,0 +1,362 @@
...
+var outputBlockSchema = &hcl.BodySchema{
+	Blocks: []hcl.BlockHeaderSchema{
+		{
+			Type:       "output",
+			LabelNames: []string{"name"},
+		},
+	},
+}
+
+func (tf TfParser) ParseServices(dir string) (map[string]bool, error) {
```
alexlokshin-czi, line 68: All of the parsing can be done in a single loop; consider for refactoring later.

shared/util/tf/tf_generator.go
alexlokshin-czi: Also, if new variables are added or removed, tf_generator will handle it, as it is both aware of module variables (including their in-depth structure), defaults (for individual object fields as well), and defaults specified in happy config. The only exception are bootstrapping vars, like `app_name` or `stack_name` -- if these variables exist on the module, we're deliberately setting their values.

cyberious:
I love the overall thought of this.

cli/pkg/stack_mgr/stack_service.go
```diff
@@ -417,3 +423,87 @@ func (s *StackService) getDistributedLock() (*backend.DistributedLock, error) {
 	lockConfig := backend.DistributedLockConfig{DynamodbTableName: s.backend.Conf().GetDynamoLocktableName()}
 	return backend.NewDistributedLock(&lockConfig, s.backend.GetDynamoDBClient())
 }
+
+func (s *StackService) Generate(ctx context.Context) error {
```
hspitzley-czi, line 427: This seems quite a bit different than the other functions in `StackService`, is there another place this could live?

alexlokshin-czi: We have workspaceRepo, orchestrator, stackService.  None of them are a perfect fit. I can create completely new tooling for it...

hspitzley-czi: I think something new would be good. As you get into modifying existing HCL I think this will grow and make StackService more multipurpose than it should be

hspitzley-czi (approved):

PR #1727, opened 2023-05-08
feat: Sidecar support for services

alexlokshin-czi:
Per Edu team request. Adds the ability to deploy sidecars alongside your application containers. To add a sidecar, include a block similar to this under your service definition:

```
    sidecars = {
        echo = {
          image  = "ealen/echo-server"
          tag    = "0.7.0"
          port   = 80
          cpu    = "100m"
          memory = "128Mi"
          image_pull_policy = "IfNotPresent" // Or "Always"
          health_check_path = "/"
          initial_delay_seconds = 30
          period_seconds = 3
        }
      }
```

Sidecar container is a secondary container, deployed alongside the primary application container, typically an agent, an initializer, a bootstrapper, a database schema migrator, etc. Sidecar container resides within the same container network as the primary application container (make sure ports don't collide), and either application can call the other one. Some sidecars can intercept traffic heading to the application (for the purposes of TLS termination, of traffic offloading). In this specific setup an an example provided sidecar responds on port 80; main application, when invoked on `/sidecar` endpoint, proxies the call to the sidecar and responds to the client.


examples/sidecar/.happy/terraform/envs/rdev/providers.tf
```diff
@@ -0,0 +1,44 @@
+# This file is autogenerated by 'happy infra generate'. Do not edit manually.
```
jakeyheath, line 1: We need to update this comment so that Github picks it up as a generated file. See the fogg comments to match. They are ignored by Github already

examples/sidecar/.happy/terraform/envs/rdev/variables.tf
```diff
@@ -0,0 +1,35 @@
...
+  type        = string
+  description = "Override the default image tags (json-encoded map)"
+  default     = "{}"
+}
+variable "stack_name" {
+  type        = string
+  description = "Happy Path stack name"
+}
+variable "wait_for_steady_state" {
+  type        = bool
```
jakeyheath, line 32: This variable is unused

examples/sidecar/README.md
```diff
@@ -0,0 +1,13 @@
+# Typical App
+
+A base typical application deployed using happy path.
```
jakeyheath, line 3: Can you update this readme?

terraform/modules/happy-service-eks/variables.tf
```diff
@@ -198,6 +198,23 @@ variable "routing" {
...
+  type = map(object({
+    image : string
+    tag : string
+    port : optional(number, 80),
+    memory : optional(string, "100Mi")
+    cpu : optional(string, "100m")
+    image_pull_policy : optional(string, "IfNotPresent")
+    health_check_path : optional(string, "/")
+    initial_delay_seconds : optional(number, 30),
+    period_seconds : optional(number, 3),
```
jakeyheath, line 212: Generally, I don't think we should really expose these to our customers. I think leaving them here as optionals is fine as long as our examples keep them out and just use the defaults. I think these fields will rarely need to be different from the defaults and makes the configuration more complicated.

terraform/modules/happy-service-eks/main.tf
```diff
@@ -173,6 +173,51 @@ resource "kubernetes_deployment_v1" "deployment" {
           }
         }
 
+        dynamic "container" {
+          for_each = coalesce(var.sidecars, {})
```
jakeyheath, line 177:
```suggestion
          for_each = var.sidecars
```

The sidecars variable has a default of `{}`, so it doesn't need the coalesce

terraform/modules/happy-service-eks/variables.tf
```diff
@@ -198,6 +198,23 @@ variable "routing" {
...
+  }))
+  default     = {}
+  description = "List of sidecars to be deployed alongside the service"
```
jakeyheath, line 215:
```suggestion
  description = "Map of sidecars to be deployed alongside the service"
```

terraform/modules/happy-service-eks/variables.tf
```diff
@@ -198,6 +198,23 @@ variable "routing" {
   })
   description = "Routing configuration for the ingress"
 }
+
+variable "sidecars" {
+  type = map(object({
+    image : string
```
jakeyheath, line 204: Sidecars are generally images that are hosted somewhere else right? They aren't images built as part of the happy project?

examples/sidecar/.happy/terraform/envs/rdev/providers.tf
alexlokshin-czi:
```
# Auto-generated by 'happy infra'. Do not edit
# Make improvements in happy, so that everyone can benefit.
```
?

terraform/modules/happy-service-eks/variables.tf
alexlokshin-czi: I disagree. Sidecars are going to be 3rd party containers with non-standard behavior. Having the ability to tweak their healthchecks is essential, as a failing sidecar will prevent the main applicaiton to start. The cerbos sidecar we're deploying will have a non-standard healthcheck, so the set of options needed to be configured will only expand.

terraform/modules/happy-service-eks/variables.tf
alexlokshin-czi: Sidecars are not built as a part of the happy pipeline. Typically, an agent of some sort.

terraform/modules/happy-service-eks/variables.tf
jakeyheath: The healthcheck one makes sense, I agree. I like the idea about tweaking them if need be, I just think we shouldn't advertise the other options. I think we'll get more questions about stuff like image_pull_policy and delay_seconds

examples/sidecar/README.md
```diff
@@ -0,0 +1,14 @@
+# A Sidecar-Enabled App
+
+A typical application deployed using happy path, that deployes a sidecar container.
```
jakeyheath, line 3: thank you! looks great

examples/sidecar/.happy/terraform/envs/rdev/providers.tf
jakeyheath: I tried to find docs on how Github knows a file is autogenerate but didn't see anything. I think this is fine

PR #1728, opened 2023-05-08
fix: Refresh EKS credentials after a lengthy docker build

alexlokshin-czi:


hspitzley-czi (approved):

shared/backend/aws/backend.go
```diff
@@ -221,21 +223,26 @@ func (b Backend) GetCredentials(ctx context.Context) (aws.Credentials, error) {
...
 		}
 	}
-	return computeBackend, nil
+	b.computeBackend = computeBackend
+	return nil
+}
+
+func (b *Backend) GetComputeBackend() compute.ComputeBackend {
+	return b.computeBackend
 }
```
jakeyheath, line 246: Why make this a Getter with a private field? Seems unnecessary

jakeyheath (approved):
The places we refresh seem arbitrary to me. I won't bug you with it now, but I think it'd be better if either refresh every time we need to use K8S or we catch an authorization error and refresh then. My other suggestion would be to only fresh after docker builds, which seem to be the longest running step.

Please add a description in your PR that includes the places you refresh now and why you are doing the refresh there.

alexlokshin-czi:
> The places we refresh seem arbitrary to me. I won't bug you with it now, but I think it'd be better if either refresh every time we need to use K8S or we catch an authorization error and refresh then. My other suggestion would be to only fresh after docker builds, which seem to be the longest running step.
> 
> Please add a description in your PR that includes the places you refresh now and why you are doing the refresh there.

Updated to refresh creds every time.

PR #1757, opened 2023-05-11
fix: session manager not needed unless using shell command

jakeyheath:


cli/cmd/build.go
```diff
@@ -28,7 +28,6 @@ var buildCmd = &cobra.Command{
 			checklist.DockerInstalled,
 			checklist.TerraformInstalled,
 			checklist.AwsInstalled,
-			checklist.AwsSessionManagerPluginInstalled,
```
alexlokshin-czi, line 31: Another improvement would be not to require session manager when running on EKS.

alexlokshin-czi (approved):

cli/cmd/build.go
jakeyheath: true

PR #1761, opened 2023-05-12
feat: Collect stack configuration from existing terraform code and store in happy config

alexlokshin-czi:
This PR enable `happy infra ingest`, that reads from existing terraform `happy-env-*` module invocations across all environments, figures out stack defaults and overrides for every environment and stores it in the happy config. This feature is limited, terraform variables or function invocations are not supported and will be ignored. To be used with the `happy infra generate` for a gradual evolution towards terraform code being generated rather than written by hand.


shared/util/maps.go
```diff
@@ -0,0 +1,168 @@
+package util
```
hspitzley-czi, line 1: What do you think about moving maps.go to a new package in the go-misc repo?

shared/hclmanager/hcl_manager.go
```diff
@@ -112,3 +112,42 @@ func (h HclManager) Generate(ctx context.Context) error {
...
+		if err != nil {
+			return errors.Wrap(err, "Unable to parse a stack module call")
+		}
+
+		moduleCall.Parameters = util.DeepCleanup(moduleCall.Parameters)
+		moduleCalls[name] = moduleCall
+		stackDefaults = moduleCall.Parameters
+	}
+
+	// Determine commpon stack defaults
```
hspitzley-czi, line 138:
```suggestion
	// Determine common stack defaults
```

hspitzley-czi, line 138:
```suggestion
	// Determine common stack defaults
```

shared/util/tf/tf_parser.go
```diff
@@ -103,3 +110,129 @@ func (tf TfParser) ParseServices(dir string) (map[string]bool, error) {
...
+	moduleCall := ModuleCall{Parameters: map[string]any{}}
+
+	err := filepath.WalkDir(dir, func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if d.IsDir() {
+			if d.Name() == ".terraform" || d.Name() == ".git" {
+				return filepath.SkipDir
+			}
```
hspitzley-czi, line 124: Why are these treated differently with SkipDir instead of nil?

shared/util/tf/tf_parser.go
```diff
@@ -103,3 +110,129 @@ func (tf TfParser) ParseServices(dir string) (map[string]bool, error) {
...
+				}
+
+				v, err := decodeValue(value)
+				if err != nil {
+					continue
+				}
+				if v != nil {
+					moduleCall.Parameters[attr.Name] = v
+				}
+			}
```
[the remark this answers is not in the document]
hspitzley-czi: Might add this as a comment in the code since it might be helpful beyond this PR

shared/util/tf/tf_parser.go
```diff
@@ -103,3 +110,129 @@ func (tf TfParser) ParseServices(dir string) (map[string]bool, error) {
...
+
+		return nil
+	})
+	if err != nil {
+		return moduleCall, errors.Wrap(err, "failed to parse terraform files")
+	}
+	return moduleCall, nil
+}
+
+func decodeValue(ctyValue cty.Value) (any, error) {
```
hspitzley-czi, line 202: How does this handle when the attr is set to a var or a local? (ie `aws_role = var.aws_role`)

shared/util/maps.go
alexlokshin-czi: Sounds like a good idea; I can do it in a followup PR.

jakeyheath (approved):

shared/util/tf/tf_parser.go
alexlokshin-czi: It will error out on such value. Currently any functional calls or variable references are not supported.

PR #1775, opened 2023-05-18
feat: Ingress for pods exposing HTTPS

ademartini-czi:
This feature allows you specify a `service_scheme` in order to support the ALB target group using HTTPS protocol.  Also fixed the private URL output so it uses the `service_port` and `service_scheme`.

Tested with this service: https://ssl-proxy-test.edu-platform.rdev.si.czi.technology/api/ping which is targeting a pod that exposes port `8443` with HTTPS.

Happy terraform for the service here: https://github.com/chanzuckerberg/edu-stack/blob/ssl-proxy-test/.happy/terraform/envs/rdev/main.tf#L59-L60.

Screenshot 2023-05-18 at 11 25 11 AM

Note - my knowledge on this ingress stuff is limited.  I've been reading here to learn: https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.5/how-it-works/#ingress-traffic.  I may be missing some details, but the above seems to work.


terraform/modules/happy-stack-eks/variables.tf
```diff
@@ -50,6 +50,7 @@ variable "services" {
     port : optional(number, 80),
     scheme : optional(string, "HTTP"),
     service_port : optional(number, null),
+    service_scheme : optional(string, "HTTP"),
```
alexlokshin-czi, line 53: We need a validation to ensure this scheme is either `HTTP` or `HTTPS`

ademartini-czi: Thanks Alex, I added validations for the three scheme fields.

jakeyheath (approved):
LGTM, thanks!

PR #1806, opened 2023-05-24
fix: [bug] Validate credentials before stack operations and prompt user to log in and create a new token on token absence or prior invalidation

alexlokshin-czi:
This PR addresses several failing scenarios related to TFE access.

* Expired token causes notifications to the user without the ability to refresh the token
* Missing token from the TFE credential file
* Token stored in credential file, but invalidated in TFE

Removed asynchronicity to allow TFE token to be validated and user prompted to login. Customer observations registered here: https://chanzuckerbergteam.slack.com/archives/C04NVSCDFR7/p1684947961570009
Screenshot 2023-05-24 at 10 05 06 AM


cli/pkg/stack_mgr/stack_service.go
```diff
@@ -342,33 +342,24 @@ func (s *StackService) CollectStackInfo(ctx context.Context, listAll bool, app s
...
-			stackInfo, err := stacks[name].GetStackInfo(ctx)
-			if err != nil {
-				log.Warnf("unable to get stack info for %s: %s (likely means the deploy failed the first time)", name, err)
-				if !diagnostics.IsInteractiveContext(ctx) {
-					stackInfos[i] = &StackInfo{
-						Name:    name,
-						Status:  "error",
-						Message: err.Error(),
-					}
+		stackInfo, err := stacks[name].GetStackInfo(ctx)
```
alexlokshin-czi, line 345: Behind the scenes, this func validates the tfe token and prompts the login. If executed in a go routine, interaction with the token refresher is not possible.

shared/workspace_repo/api_client.go
```diff
@@ -87,7 +87,7 @@ func (c *WorkspaceRepo) getToken(hostname string) (string, error) {
 }
 
 func (c *WorkspaceRepo) getTfc(ctx context.Context) (*tfe.Client, error) {
-	if c.tfc == nil {
+	if c.tfc == nil || c.validateAccess(ctx) != nil {
```
alexlokshin-czi, line 90: Check if the token is still valid.

shared/workspace_repo/api_client.go
```diff
@@ -172,7 +178,7 @@ func (c *WorkspaceRepo) enforceClient(ctx context.Context) (*tfe.Client, error)
 					return nil, errors.Wrap(err, "provided tfe token not valid, and cannot refresh it in a non-interactive mode")
 				}
 				errs = multierror.Append(errs, err)
-				state = tokenRefreshNeeded
+				state = tokenMissing
```
alexlokshin-czi, line 181: Token refresh only refreshes an existing token by logging user into TFE; if the token was previously invalidated, it will just keep prompting to open the browser window. Setting the state to tokenMissing if token still doesn't work will force tfeLogin().

shared/workspace_repo/api_client.go
```diff
@@ -106,6 +106,11 @@ func (c *WorkspaceRepo) getTfc(ctx context.Context) (*tfe.Client, error) {
 	return c.tfc, nil
 }
 
+func (c *WorkspaceRepo) validateAccess(ctx context.Context) error {
+	_, err := c.tfc.Organizations.List(ctx, &tfe.OrganizationListOptions{})
```
jakeyheath, line 110: Does this throw an error when the token is invalid? Looking at the source code for this library, it seems like it only returns the errors of `req.Do`  https://github.com/hashicorp/go-tfe/blob/b5d742f97b1ba8f474da7d467df1566c2aee4852/organization.go#LL254C13-L254C13. According to the docs, req.Do doesn't return an error for non-200 HTTP responses.

cli/pkg/stack_mgr/stack_service.go
jakeyheath: sounds like the enforceClient is not threadsafe. I made this a concurrent operation because GetStackInfo is kind of a slow operation and when lots of stacks exist, it can be pretty time consuming to query all the info. I think if we make the enforceClient function thread safe, I would prefer to keep this in a goroutine. I think doing so will also open the door to using more concurrency later too. Let me experiment so I can provide an alternate solution for you review

alexlokshin-czi: enforceClient is not threadsafe, as it interacts with the stdin.

alexlokshin-czi: The only way to work around it is to enforce the client before the TFE operations, but it's not always a linear process as TFE interactions are hidden behind layers of indirection.

shared/workspace_repo/api_client.go
alexlokshin-czi: It will throw an error if the token is missing or invalid -- went through pretty thorough testing.

cli/pkg/stack_mgr/stack_service.go
jakeyheath: https://github.com/chanzuckerberg/happy/pull/1814

jakeyheath (approved):

PR #1814, opened 2023-05-30
feat: add back in concurrency

jakeyheath:


shared/workspace_repo/api_client.go
```diff
@@ -86,27 +88,73 @@ func (c *WorkspaceRepo) getToken(hostname string) (string, error) {
...
+	client := <-TFCClientChan
+	err := <-ErrorChan
+	return client, err
+}
+
+func (c *WorkspaceRepo) validateAccessSince(ctx context.Context, validatedSince time.Duration) error {
+	if time.Since(c.lastValidated) < validatedSince {
+		return nil
+	}
+	c.lastValidated = time.Now()
```
jakeyheath, line 157: Only validate every 10 minutes, not every time.

shared/workspace_repo/api_client.go
```diff
@@ -192,7 +240,7 @@ func (c *WorkspaceRepo) Stacks() ([]string, error) {
 }
 
 func (c *WorkspaceRepo) GetWorkspace(ctx context.Context, workspaceName string) (Workspace, error) {
-	client, err := c.getTfc(ctx)
+	client, err := c.getTfcSync()
```
jakeyheath, line 243: This operation needs to be threadsafe so multiple goroutines can get the same client and validation only needs to happen one time. A new function was made to indicate it is a synchronous operation.

shared/workspace_repo/api_client.go
```diff
@@ -86,27 +88,73 @@ func (c *WorkspaceRepo) getToken(hostname string) (string, error) {
...
 
-func (c *WorkspaceRepo) validateAccess(ctx context.Context) error {
+// Thread-safe version of getTfc . Needed because the credential
+// check that happens on all TFC clients needs to be used by multiple goroutines
+// without triggering the credential check multiple times.
+func (c *WorkspaceRepo) getTfcSync() (*tfe.Client, error) {
+	WorkspaceRepoChan <- c
+	client := <-TFCClientChan
+	err := <-ErrorChan
+	return client, err
```
jakeyheath, line 150: Utilize the worker pool to get the client. This helps use take care of thread safety because we only have 1 worker in this worker pool.

alexlokshin-czi (approved):

PR #1824, opened 2023-05-31
feat: Implement "happy infra validate" to validate terraform scripts

alexlokshin-czi:
This feature improves `happy infra ingest` and adds `happy infra validate` that examines the values configured in a `happy-stack-*` modules calls for type compatibility. If it picks up on member type inconsistencies, it will notify the user with messages like:

```
[WARNING]: Provided value for attribute 'services' doesn't match the one required by the module: type mismatch for member frontend: type mismatch for attribute port: expected: number, got: string
[WARNING]: Provided value for attribute 'routing_method' doesn't match the one required by the module: expected: string, got: number
```


jakeyheath (approved):
isn't this done by terraform? If I ran `terraform validate` I think you'd get these same errors. Also, the terraform plan will through these errors too right?

alexlokshin-czi:
> isn't this done by terraform? If I ran `terraform validate` I think you'd get these same errors. Also, the terraform plan will through these errors too right?

Yes, this is done by terraform as well, but this gives a much deeper insight into the structural differences. Key reason for this feature to exist is to make sure `happy ingest` consumes valid configurations.

PR #1825, opened 2023-05-31
fix: deadlock in deletes

jakeyheath:
This PR fixes a deadlock in deletes because the worker pool isn't started.

### DIFF
--- a/cli/cmd/delete.go
+++ b/cli/cmd/delete.go
@@ -3,12 +3,10 @@ package cmd
 import (
 	"context"
 	"fmt"
-	"time"
 
 	"github.com/AlecAivazis/survey/v2"
-	"github.com/chanzuckerberg/happy/cli/pkg/cmd"
+	happyCmd "github.com/chanzuckerberg/happy/cli/pkg/cmd"
 	"github.com/chanzuckerberg/happy/cli/pkg/orchestrator"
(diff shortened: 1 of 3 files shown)


cli/cmd/delete.go
```diff
@@ -44,59 +45,48 @@ var deleteCmd = &cobra.Command{
 }
 
 func runDelete(cmd *cobra.Command, args []string) error {
-	ctx := context.WithValue(cmd.Context(), options.DryRunKey, dryRun)
 	stackName := args[0]
 
-	bootstrapConfig, err := config.NewBootstrapConfig(cmd)
+	happyClient, err := makeHappyClient(cmd, sliceName, stackName, []string{tag}, createTag)
```
alexlokshin-czi, line 50: I traced the code to starting the worker pool, and see that `config.NewHappyConfig()` is called in a bunch of other places --  does it make sense to make the tfc worker pool global and create it globally so we won't miss another one of these locks? Or, perhaps, not synchronize if there's no pool?

jakeyheath: the only reason I did it there was because we need a context. Maybe I can find a place to grab the context for all commands

jakeyheath: see https://github.com/chanzuckerberg/happy/pull/1825/commits/08a6167162357198fcd8fdd5930438536529ff8d @alexlokshin-czi

alexlokshin-czi (approved):

PR #1841, opened 2023-06-02
fix: Gracefully fail if module invocation doesn't have the source specified

alexlokshin-czi:
Fixes [CCIE-1515]. When running against a folder of HCL code with no terraform module invocations, `happy infra validate` would fail because it cannot locate the `source` attribute.

[CCIE-1515]: https://czi-tech.atlassian.net/browse/CCIE-1515?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/api-ci.yml
```diff
@@ -39,10 +39,9 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v3
         with:
-          version: latest
+          version: v1.52.2
```
alexlokshin-czi, line 42: 1.53 has a bug, had to pin to 1.52.

hspitzley-czi (approved):

PR #1842, opened 2023-06-02
feat: Add sidecar support to happy shell

alexlokshin-czi:


shared/backend/aws/k8s_compute_backend.go
```diff
@@ -304,11 +304,19 @@ func (k8s *K8SComputeBackend) Shell(ctx context.Context, stackName string, servi
...
-	if len(pod.Spec.Containers) > 1 {
-		return errors.Errorf("There's more than one container in a pod '%s'", podName)
+	if len(pod.Spec.Containers) > 1 && len(containerName) == 0 {
+		containerNames := []string{}
+		for _, container := range pod.Spec.Containers {
+			containerNames = append(containerNames, container.Name)
+		}
+
+		logrus.Warnf("There's more than one container in a pod '%s': %s", podName, strings.Join(containerNames, ", "))
+		return errors.New("Please specify container name via --container flag")
```
jakeyheath, line 314: instead of an error, maybe a prompt would be better here since most people familiar with this feature won't be used to typing --container yet

alexlokshin-czi: That's a thought.

jakeyheath (approved):

PR #1867, opened 2023-06-09
feat: consolidate stack operations on cli and hapi

jakeyheath:
This PR updates the happy list API functionality with the exact same function calls that are utilized in the CLI. It also adds code that can create happy helper functions that the CLI uses passed on parameters passed to the API request.


api/pkg/cmd/stack.go
```diff
@@ -22,80 +16,35 @@ type StackManager interface {
...
-
-	return stackInfos, nil
-}
-
-func getStackInfo(ctx context.Context, payload model.AppStackPayload, stackName string, workspaceRepo workspace_repo.WorkspaceRepoIface) (*model.AppStackResponse, error) {
-	stack := &model.AppStackResponse{
-		AppMetadata: *model.NewAppMetadata(payload.AppName, payload.Environment, stackName),
-	}
-	workspace, err := workspaceRepo.GetWorkspace(ctx, fmt.Sprintf("%s-%s", payload.AppMetadata.Environment, stackName))
+	happyClient, err := utils.MakeHappyClient(ctx, payload.AppMetadata)
```
jakeyheath, line 33: we not have a `MakeHappyClient` that is implemented in the CLI so that the API requests can have a happy client based on the query parameters passed from the user.

api/pkg/utils/utils.go
```diff
@@ -13,3 +21,33 @@ func StructToMap(payload interface{}) (map[string]interface{}, error) {
...
+	}
+
+	// TODO: we need to enforce the TFE_TOKEN exists on boot
+	// TODO: do we want to pass in the TFE_TOKEN instead of building with a TFE_TOKEN as part of the environment?
+	tfeToken := os.Getenv("TFE_TOKEN")
+	if tfeToken == "" {
+		return nil, errors.New("TFE_TOKEN is required; this should never happen")
+	}
+	workspaceRepo := workspace_repo.NewWorkspaceRepo(awsBackend.Conf().GetTfeUrl(), awsBackend.Conf().GetTfeOrg()).
+		WithTFEToken(tfeToken)
```
jakeyheath, line 43: use the TFE_TOKEn from the environment. Maybe the user should pass this in?

shared/model/app_stack.go
```diff
@@ -1,27 +1,70 @@
...
+	"github.com/chanzuckerberg/happy/shared/k8s"
+	"github.com/chanzuckerberg/happy/shared/util"
+)
+
 type AppStack struct {
 	CommonDBFields
 	AppMetadata // TODO: might want to change this to AppStackPayload but going with minimal columns for now
 }
 
+type StackMetadata struct {
```
jakeyheath, line 17: copied from the CLI

shared/stack/stack_meta.go
```diff
@@ -0,0 +1,269 @@
+package stack
```
jakeyheath, line 1: same as stack.go (copied from the CLI exactly)

api/pkg/cmd/stack.go
```diff
@@ -22,80 +16,41 @@ type StackManager interface {
...
-		}(i, stackName)
-	}
-	wg.Wait()
-
-	return stackInfos, nil
-}
-
-func getStackInfo(ctx context.Context, payload model.AppStackPayload, stackName string, workspaceRepo workspace_repo.WorkspaceRepoIface) (*model.AppStackResponse, error) {
-	stack := &model.AppStackResponse{
-		AppMetadata: *model.NewAppMetadata(payload.AppName, payload.Environment, stackName),
```
jakeyheath, line 85: removed reimplementation of CollectStackInfo

shared/stack/stack_service.go
```diff
@@ -0,0 +1,430 @@
...
+func NewStackService(env, appName string) *StackService {
+	return &StackService{
+		executor: util.NewDefaultExecutor(),
+		env:      env,
+		appName:  appName,
+	}
+}
+
+func (s *StackService) GetWritePath() string {
+	return fmt.Sprintf("/happy/%s/stacklist", s.env)
```
alexlokshin-czi, line 54: This needs to be pushed down to the implementation - k8s_compute_backend or ecs_compute_backend.

shared/stack/stack_service.go
```diff
@@ -0,0 +1,430 @@
...
+}
+
+func (s *StackService) GetNamespacedWritePath() string {
+	return fmt.Sprintf("/happy/%s/%s/stacklist", s.appName, s.env)
```
alexlokshin-czi, line 58: Same as above.

shared/stack/stack_service.go
```diff
@@ -0,0 +1,430 @@
...
+	cb, err := s.backend.GetComputeBackend(ctx)
+	if err != nil {
+		return errors.Wrap(err, "unable to connect to a compute backend")
+	}
+
+	if err := cb.WriteParam(ctx, s.GetNamespacedWritePath(), stackNamesStr); err != nil {
+		return errors.Wrap(err, "unable to write a workspace param")
+	}
+	log.WithFields(log.Fields{"path": s.GetWritePath(), "data": stackNamesStr}).Debug("Writing to paramstore...")
+	if err := cb.WriteParam(ctx, s.GetWritePath(), stackNamesStr); err != nil {
```
alexlokshin-czi, line 288: Which one of these two params is the source of truth?

shared/stack/stack_service.go
```diff
@@ -0,0 +1,430 @@
...
+}
+
+func (s *StackService) HasState(ctx context.Context, stackName string) (bool, error) {
+	workspace, err := s.GetStackWorkspace(ctx, stackName)
+	if err != nil {
+		if errors.Is(err, tfe.ErrInvalidWorkspaceValue) || errors.Is(err, tfe.ErrResourceNotFound) {
+			// Workspace doesn't exist, thus no state
+			return false, nil
+		}
+		return true, errors.Wrap(err, "Cannot get the stack workspace")
```
alexlokshin-czi, line 422:
```suggestion
		return true, errors.Wrap(err, "cannot get the stack workspace")
```

shared/stack/stack_service.go
jakeyheath: all this code is copied from the CLI. Do you want me to reimplement it in this PR?

shared/stack/stack_service.go
jakeyheath: agreed, but I copied this code from the CLI so that it could be leverage by the API. Should I fix this now or should I make another PR to fix it?

api/pkg/request/happy_client.go
```diff
@@ -0,0 +1,41 @@
...
+	AWSBackend   *backend.Backend
+}
+
+func MakeHappyClient(ctx context.Context, appName string, envCtx config.EnvironmentContext) (*HappyClient, error) {
+	awsBackend, err := backend.NewAWSBackend(ctx, envCtx,
+		backend.WithNewAWSConfigOption(configv2.WithRegion(*envCtx.AWSRegion)),
+		backend.WithNewAWSConfigOption(configv2.WithCredentialsProvider(MakeCredentialProvider(ctx))),
+	)
+	if err != nil {
+		return nil, errors.Wrap(err, "making AWS backend")
```
alexlokshin-czi, line 26:
```suggestion
		return nil, errors.Wrap(err, "failed to construct an AWS backend")
```

shared/backend/aws/backend.go
```diff
@@ -182,15 +186,15 @@ func NewAWSBackend(
 
 	userName, err := b.GetUserName(ctx)
 	if err != nil {
-		return nil, errors.Wrapf(err, "unable to retrieve identity info, does aws profile [%s] exist?", *b.awsProfile)
+		return nil, errors.Wrapf(err, "retrieving identity info")
```
alexlokshin-czi, line 189: Why?

shared/stack/stack.go
```diff
@@ -0,0 +1,368 @@
...
+		}
+
+		cmdPath, err := s.executor.LookPath("tflocal")
+		if err != nil {
+			return errors.Wrap(err, "failed to locate tflocal")
+		}
+
+		// Clear out any prior state... For now. Every stack has to have its own
+
+		// _ = os.Remove(filepath.Join(srcDir, "terraform.tfstate"))
```
alexlokshin-czi, line 219: Let's remove these comments

api/pkg/cmd/stack.go
```diff
@@ -22,80 +16,41 @@ type StackManager interface {
...
-	workspace, err := workspaceRepo.GetWorkspace(ctx, fmt.Sprintf("%s-%s", payload.AppMetadata.Environment, stackName))
+	// we cancel the context to close up any spun up goroutines
+	// for this threads workspace_repo
+	// TODO: we should probably cache these credentials/clients to TFE
+	ctx, done := context.WithCancel(ctx)
+	defer done()
+	happyClient, err := request.MakeHappyClient(ctx, payload.AppName, payload.MakeEnvironmentContext(payload.Environment))
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to get workspace")
+		return nil, errors.Wrap(err, "making happy client")
```
alexlokshin-czi, line 40:
```suggestion
		return nil, errors.Wrap(err, "failed to create a happy client")
```

api/pkg/cmd/stack.go
```diff
@@ -22,80 +16,41 @@ type StackManager interface {
...
 	}
 
-	stack.WorkspaceUrl = workspace.GetWorkspaceUrl()
-	stack.WorkspaceStatus = workspace.GetCurrentRunStatus(ctx)
-	stack.WorkspaceRunUrl = workspace.GetCurrentRunUrl(ctx)
-	stack.Endpoints, err = workspace.GetEndpoints(ctx)
+	stacks, err := happyClient.StackService.CollectStackInfo(ctx, false, payload.AppName)
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to get stack endpoints")
+		return nil, errors.Wrapf(err, "collecting stack info")
```
alexlokshin-czi, line 45:
```suggestion
		return nil, errors.Wrapf(err, "failed to collect stack info")
```

api/pkg/cmd/stack.go
jakeyheath: actually I just learned that we shouldn't be adding "failed to" or "errored at"

jakeyheath: https://github.com/uber-go/guide/blob/master/style.md#error-wrapping

api/pkg/cmd/stack.go
jakeyheath: https://github.com/uber-go/guide/blob/master/style.md#error-wrapping

shared/backend/aws/backend.go
jakeyheath: in some cases, awsProfile is nil and it was causing a panic (the API doesn't use an aws profile)

api/pkg/cmd/stack.go
alexlokshin-czi: I don't think that makes it more readable. It obscures the nature of the issue.

api/docs/swagger.json
```diff
@@ -502,18 +505,51 @@
...
+                },
+                "tfe_workspace_run_url": {
                     "type": "string"
                 },
-                "workspace_status": {
+                "tfe_workspace_status": {
                     "type": "string"
                 },
-                "workspace_url": {
+                "tfe_workspace_url": {
```
hspitzley-czi, line 552: This is a breaking change for eng-portal, let's coordinate the release

shared/model/app_stack.go
```diff
@@ -1,27 +1,71 @@
...
+	Name               string            `json:"name,omitempty"`
```
hspitzley-czi, line 17: Is this the stack name? If so, we will have a duplicate field in `AppStackResponse` because `AppMetadata` has the stack name too, maybe we can find a way to consolidate

shared/model/app_stack.go
```diff
@@ -1,27 +1,71 @@
...
+	Owner              string            `json:"owner,omitempty"`
+	Tag                string            `json:"tag,omitempty"`
+	Status             string            `json:"status,omitempty"`
+	LastUpdated        string            `json:"last_updated,omitempty"`
+	Message            string            `json:"message,omitempty"`
+	Outputs            map[string]string `json:"outputs,omitempty"`
+	Endpoints          map[string]string `json:"endpoints,omitempty"`
+	App                string            `json:"app,omitempty"`
```
hspitzley-czi, line 25: We will have a duplicate field from this as well because `AppMetadata` has the app name too, maybe we can find a way to consolidate

api/docs/swagger.json
hspitzley-czi: Up to you, I can make eng-portal accept either if you do want to do the rename

PR #1870, opened 2023-06-12
fix: messaging around slice validation

jakeyheath:
This PR messed up the error messaging about where the problem is with the configuration, which leads to confusing when trying to solve it.


cli/cmd/init.go
```diff
@@ -246,7 +246,7 @@ func validateConfigurationIntegirty(ctx context.Context, slice string, happyClie
 			}
 			ok := ss.ContainsElement(serviceName)
 			if !ok {
-				return errors.Errorf("service %s is not configured in docker-compose.yml, but referenced in .happy/config.json", serviceName)
+				return errors.Errorf("service %s was referenced in docker-compose.yml as part of this slice, but not referenced in .happy/config.json services array", serviceName)
```
alexlokshin-czi, line 249: This is not a slice specific error; it could be confusing when slice functionality is not used.
```suggestion
				return errors.Errorf("service %s was referenced in docker-compose.yml, but not referenced in .happy/config.json services array", serviceName)
```

PR #1949, opened 2023-06-20
fix: deadlock in API by not starting worker pool

jakeyheath:
The workspace information is gathered by goroutines, so the credential check is preferred by a single worker. However, the worker pool wasn't started in the API causing the channel to block and return 504 timeout errors on calls to the stacklist API.

### DIFF
--- a/api/pkg/cmd/stack.go
+++ b/api/pkg/cmd/stack.go
@@ -60,6 +60,10 @@ func enrichStacklistMetadata(ctx context.Context, stacklist []string, payload mo
 		integrationSecret.Tfe.Url,
 		integrationSecret.Tfe.Org,
 	).WithTFEToken(setup.GetConfiguration().TFE.Token)
+	ctx, done := context.WithCancel(ctx)
+	defer done()
+	workspace_repo.StartTFCWorkerPool(ctx)
+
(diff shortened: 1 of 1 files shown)


api/pkg/cmd/stack.go
```diff
@@ -60,6 +60,10 @@ func enrichStacklistMetadata(ctx context.Context, stacklist []string, payload mo
 		integrationSecret.Tfe.Url,
 		integrationSecret.Tfe.Org,
 	).WithTFEToken(setup.GetConfiguration().TFE.Token)
+	ctx, done := context.WithCancel(ctx)
+	defer done()
+	workspace_repo.StartTFCWorkerPool(ctx)
```
alexlokshin-czi, line 65: Perhaps, we should start it globally in one place?

jakeyheath: I'm about to refactor this, I just want to fix it while I am working on the refactor. The tough part about starting it globally is I need a context object so I can kill the pool. Also the concurrency model is a little different in the API than the CLI, so this is a little bit of a cover while I rework how the concurrency works.

hspitzley-czi (approved):

PR #1956, opened 2023-06-21
fix: happy infra ingest fails for the tasks example

alexlokshin-czi:
Fixes CCIE-1527. Interpolations in task references cause ingestion issues. Also fixed generation aspect of it, with interpolations present in the output. Main benefit of this PR is the ability to reference terraform variables in config.json, ingest terraform code that references tf variables in sub-values, and generate terraform code that references variables where it was not previously possible.


shared/util/tf/tf_parser.go
```diff
@@ -223,7 +227,18 @@ func (tf TfParser) ParseModuleCall(dir string) (ModuleCall, error) {
 					continue
 				}
 
-				value, diag := attr.Expr.Value(nil)
+				inputVariables := map[string]cty.Value{}
+				for _, variable := range []string{"image_tag", "image_tags", "k8s_namespace", "stack_name", "deployment_stage", "stack_prefix"} {
```
alexlokshin-czi, line 231: A list of variables that can be referenced in attribute interpolations.

jakeyheath (approved):

PR #1960, opened 2023-06-21
feat: reuse happy client

jakeyheath:
This PR fixes a bug in the CLI where we were reauthenticating to AWS and EKS every time we talked to the API or EKS. Sometimes, this was happening 3 or 4 times on a single CLI invocation. This solves the issue where checking the API version is slow. To solve this, I cached the happy client and also passed through the AWS backend to all our structures.


cli/cmd/api.go
```diff
@@ -32,18 +32,17 @@ var apiHealthCmd = &cobra.Command{
 	Long:         "ping the health endpoint of happy api",
 	SilenceUsage: true,
 	RunE: func(cmd *cobra.Command, args []string) error {
-		happyConfig, err := config.GetHappyConfigForCmd(cmd)
+		happyClient, err := makeHappyClient(cmd, sliceName, "", []string{}, false)
```
alexlokshin-czi, line 35: Might be useful to introduce an options arg with default values

alexlokshin-czi: It looks like happyClient is always created with the same exact arguments: 
```go
makeHappyClient(cmd, sliceName, "", []string{}, false)
```

cli/cmd/api.go
```diff
@@ -32,18 +32,17 @@ var apiHealthCmd = &cobra.Command{
...
 		if err != nil {
 			return err
 		}
-
-		api := hapi.MakeApiClient(happyConfig)
+		api := hapi.MakeAPIClient(happyClient.HappyConfig, happyClient.AWSBackend)
```
alexlokshin-czi, line 39: Can MakeAPIClient be made dependent on happyClient alone? This will simplify signatures -- also I doubt there's an intention to use happyConfig and AWSBackend coming from different sources.

cli/cmd/api.go
jakeyheath: yea I can cut a new PR to refactor this function

cli/cmd/api.go
jakeyheath: we make API clients in other contexts, for example the terraform provider where the happy Client won't exist

alexlokshin-czi: I see.

alexlokshin-czi (approved):

PR #1982, opened 2023-06-28
fix: consolidate data structures and use new shared packages

jakeyheath:
This PR updates some of the shared functionality to utilize model packages for responses and removes duplicate fields from the CLI and the API.


hspitzley-czi (approved):
:shipit:

PR #1983, opened 2023-06-28
chore: remove module instance-cloud-init-script

naihsuanshao:
Remove the module instance-cloud-init-script in batch.tf and make creating cloud init script optional.


jakeyheath:
stale

PR #2066, opened 2023-07-19
feat: CCIE-1619: happy version manager v1

dspadea:
This PR:

* Creates the first cut of the `hvm` version manager
* Makes the `happy` repo into a Go workspace to allow modules to depend on each other without needing to go through Github to resolve new code
* Changes the format of the `.happy/versions.lock` file to support arbitrary lists of packages and versions

<details open>
  <summary><a href="https://czi-tech.atlassian.net/browse/CCIE-1619" title="CCIE-1619" target="_blank">CCIE-1619</a></summary>
  <br />
  <table>
    <tr>
      <th>Summary</th>
  <td>Finish up v1 of Happy Version Manager</td>
    </tr>
    <tr>
      <th>Type</th>
      <td>
        Task
        Task
      </td>
    </tr>
    <tr>
      <th>Status</th>
      <td>In Progress</td>
    </tr>
  </table>
</details>

---

null


hvm/cmd/install.go
```diff
@@ -0,0 +1,73 @@
+/*
+Copyright © 2023 NAME HERE <EMAIL ADDRESS>
```
alexlokshin-czi, line 2:
```suggestion
```

dspadea: Ugh. Yeah. Those keep getting added by the Cobra cli tool. I'll remove them. I got a bunch, but guess I missed a few. Thanks.

hvm/linkManager/set_bin_link.go
```diff
@@ -0,0 +1,72 @@
...
+
+		info, err := file.Info()
+		if err != nil {
+			return errors.Wrapf(err, "Error getting file info for %s", file.Name())
+		}
+
+		fmt.Println("Checking ", file.Name(), " permissions", info.Mode())
+		// Skip if the file is not owner-executable
+		if !strings.Contains(info.Mode().String(), "x") {
+			fmt.Println("Skipping ", file.Name(), " as it is not executable")
```
alexlokshin-czi, line 54: Let's use logrus so we can control logging levels
```suggestion
			logrus.Println("Skipping ", file.Name(), " as it is not executable")
```

dspadea:
Gotta figure out what's going on with those lints, but any other comments would be appreciated. (Or suggestions on how to fix the lints!)

alexlokshin-czi:
> Gotta figure out what's going on with those lints, but any other comments would be appreciated. (Or suggestions on how to fix the lints!)

Lints are weird. Our GH workflows use 1.52, brew installs 1.53, and there's a breaking change, if I recall correctly. You might want to install 1.52 and run it locally: `go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.52.2` and `golangci-lint run`. Provider linter fails with

```
ERRO [linters_context] typechecking error: pattern ./...: directory prefix . does not contain modules listed in go.work or their selected dependencies 
```

go.work
```diff
@@ -0,0 +1,9 @@
+go 1.20
```
jakeyheath, line 1: o wow! thanks for adding this!

hvm/cmd/download.go
```diff
@@ -0,0 +1,55 @@
...
+	org := args[0]
+	project := args[1]
+	version := args[2]
+
+	os := runtime.GOOS
+	arch := runtime.GOARCH
+	path := "."
+
+	if cmd.Flags().Changed("os") {
+		os = cmd.Flag("os").Value.String()
```
jakeyheath, line 41:
```suggestion
		os = cmd.Flags().GetString("os")
```

I think this is pretty much exactly the same, but how its documented to grab the value from a flag

hvm/cmd/download.go
```diff
@@ -0,0 +1,55 @@
...
+
+func downloadPackage(cmd *cobra.Command, args []string) {
+
+	org := args[0]
+	project := args[1]
+	version := args[2]
+
+	os := runtime.GOOS
+	arch := runtime.GOARCH
+	path := "."
```
jakeyheath, line 38: defaults can be set above in line 25-27, so like this:

~~~go
downloadCmd.Flags().StringP("arch", "a", runtime.GOARCH, "Force architecture (Default: current)")
~~~

jakeyheath: then you don't need all the conditionals below

hvm/cmd/env.go
```diff
@@ -0,0 +1,116 @@
...
+
+				}
+
+				versionPaths = append(versionPaths, swPath)
+			}
+			managedPath = strings.Join(versionPaths, ":")
+		} else {
+			fmt.Printf("export PATH=%s", basePath)
+			return
+		}
```
jakeyheath, line 96: refactor to handle the else statement first:

~~~
if !config.DoesHappyVersionLockFileExist(projectRoot) { 
  fmt.Printf("export PATH=%s", basePath)
  return
}
~~~

these reduces the amount of nesting for the base case

https://github.com/uber-go/guide/blob/master/style.md#reduce-nesting

hvm/cmd/install.go
```diff
@@ -0,0 +1,72 @@
...
+
+}
+
+func installPackage(cmd *cobra.Command, args []string) {
+
+	org := args[0]
+	project := args[1]
+	version := args[2]
+	os := runtime.GOOS
+	arch := runtime.GOARCH
```
jakeyheath, line 40: see above comment about default arguments

hvm/cmd/install.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/chanzuckerberg/happy/hvm/installer"
+	"github.com/spf13/cobra"
+)
+
+// installCmd represents the install command
+var installCmd = &cobra.Command{
+	Use:   "install [org] [project] [version]",
+	Short: "Install a version of Happy",
+	Long:  `Install a version of Happy to ~/.happy/versions/ and set it as the current version.`,
+	Run:   installPackage,
```
jakeyheath, line 20:
```suggestion
	RunE:   installPackage,
```

capture the error

hvm/cmd/install.go
```diff
@@ -0,0 +1,72 @@
...
+
+	if cmd.Flags().Changed("arch") {
+		arch = cmd.Flag("arch").Value.String()
+	}
+
+	user, err := user.Current()
+
+	if err != nil {
+		fmt.Println("Error getting current user information", err)
+		return
```
jakeyheath, line 54: when you refactor to `RunE`, return these errors with `errors.Wrap(err, "getting current user information"`

hvm/cmd/install.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	if cmd.Flags().Changed("arch") {
+		arch = cmd.Flag("arch").Value.String()
+	}
+
+	user, err := user.Current()
+
+	if err != nil {
+		fmt.Println("Error getting current user information", err)
```
jakeyheath, line 53: avoid using fmt Println, cobra will print the error messages when they are returned with `RunE`

hvm/cmd/list-releases.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/chanzuckerberg/happy/shared/githubconnector"
+	"github.com/spf13/cobra"
+)
+
+// useCmd represents the use command
+var listRelasesCommand = &cobra.Command{
+	Use: "list-releases [org] [project]",
+
+	Short: "Get list of available releases",
+	Long:  ``, Run: func(cmd *cobra.Command, args []string) {
```
jakeyheath, line 15:
```suggestion
	Long:  ``, 
	RunE: func(cmd *cobra.Command, args []string) {
```

hvm/cmd/list-releases.go
```diff
@@ -0,0 +1,46 @@
...
+		fmt.Println("use called")
+	},
```
jakeyheath, line 17: debugging leftover?

hvm/cmd/list-releases.go
```diff
@@ -0,0 +1,46 @@
...
+func listReleases(cmd *cobra.Command, args []string) error {
+
+	org := args[0]
+	project := args[1]
+
+	client := githubconnector.NewConnectorClient()
+	releases, err := client.GetReleases(org, project)
+
+	if err != nil {
+		fmt.Println("An error occurred getting the release list: ", err)
```
jakeyheath, line 37:
```suggestion
    return errors.Wrap(err, "getting release list")
```

avoid using fmt, the error returned will be printed.

hvm/cmd/lock.go
```diff
@@ -0,0 +1,66 @@
...
+	org := args[0]
+	project := args[1]
+	version := args[2]
+
+	var lockfile *config.HappyVersionLockFile
+
+	happyConfig, err := config.GetHappyConfigForCmd(cmd)
+	if err != nil {
+		fmt.Println("Error getting happy config: ", err)
+		return
```
jakeyheath, line 35: same comment about returning errors here

hvm/cmd/lock.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/chanzuckerberg/happy/shared/config"
+	"github.com/spf13/cobra"
+)
+
+// lockCmd represents the lock command
+var lockCmd = &cobra.Command{
+	Use:   "lock [org] [project] [version]",
+	Short: "Lock the current version of happy in the current project",
+	Long:  `Lock the current version of happy in the current project. This will create a .happy/version.lock file`,
+	Run:   setLock,
```
jakeyheath, line 15:
```suggestion
	RunE:   setLock,
```

hvm/cmd/lock.go
```diff
@@ -0,0 +1,66 @@
...
+
+	if !config.DoesHappyVersionLockFileExist(projectRoot) {
+		lockfile, err = config.NewHappyVersionLockFile(projectRoot)
+	} else {
+		lockfile, err = config.LoadHappyVersionLockFile(projectRoot)
+		if err != nil {
+			fmt.Println("Error loading version lockfile: ", err)
+			return
+		}
+	}
```
jakeyheath, line 48:
```suggestion
        lockfile, err = config.NewHappyVersionLockFile(projectRoot)
	if config.DoesHappyVersionLockFileExist(projectRoot) {
		lockfile, err = config.LoadHappyVersionLockFile(projectRoot)
		if err != nil {
			return errors.Wrap(err, "loading version lockfile")
	}
```

jakeyheath: - https://github.com/uber-go/guide/blob/master/style.md#reduce-nesting
- https://github.com/uber-go/guide/blob/master/style.md#error-wrapping

hvm/cmd/root.go
```diff
@@ -0,0 +1,38 @@
...
+	// has an action associated with it:
+	// Run: func(cmd *cobra.Command, args []string) { },
+}
+
+// Execute adds all child commands to the root command and sets flags appropriately.
+// This is called by main.main(). It only needs to happen once to the rootCmd.
+func Execute() {
+	err := rootCmd.Execute()
+	if err != nil {
+		os.Exit(1)
```
jakeyheath, line 24: print the error maybe instead of silently exit

```suggestion
		logrus.Fatal(err)
```

hvm/cmd/set-default.go
```diff
@@ -0,0 +1,36 @@
...
+)
+
+var setDefaultCmd = &cobra.Command{
+	Use:   "set-default [org] [project] [version]",
+	Short: "Symlink the specified version of happy to $HOME/.czi/bin to be used as default",
+	Long: `Create a symbolic link $HOME/.czi/bin/ pointing to the specified version of happy. Assuming
+$HOME/.czi/bin is set appropriately in your $PATH, this version will be used by default when running 'happy'
+outside of a project, or when a happy version config is not present.
+	`,
+	Run: setDefaultVersion,
```
jakeyheath, line 17:
```suggestion
	RunE: setDefaultVersion,
```

hvm/installer/install.go
```diff
@@ -0,0 +1,49 @@
...
+	err = doInstall(downloaded, binPath)
+
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func doInstall(sourcePackagePath, binPath string) error {
```
jakeyheath, line 33:
```suggestion
func doInstall(ctx context.Context, sourcePackagePath, binPath string) error {
```

thread through the context rather than using context.TODO. Cobra will have a context object and you can pass it in using `cmd.Context()`. It probably doesn't help too much, but will be more important if this ever uses goroutines in the future.

jakeyheath: > Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx:

https://pkg.go.dev/context

hvm/installer/install.go
```diff
@@ -0,0 +1,49 @@
...
+
+	data, _ := ioutil.ReadFile(sourcePackagePath)
+	buffer := bytes.NewBuffer(data)
+	err = extract.Gz(context.TODO(), buffer, binPath, nil)
+
+	if err != nil {
+		return errors.Wrapf(err, "Error extracting package %s", sourcePackagePath)
+	}
+
+	return nil
```
jakeyheath, line 48:
```suggestion
	return errors.Wrapf(extract.Gz(ctx, buffer, binPath, nil), "extracting package %s", sourcePackagePath)
```

hvm/linkManager/set_bin_link.go
```diff
@@ -0,0 +1,72 @@
...
+func SetBinLink(org, project, version string) error {
+
+	user, err := user.Current()
+
+	if err != nil {
+		return errors.Wrapf(err, "Error getting current user information")
+	}
+
+	home := user.HomeDir
+	versionsPath := path.Join(home, ".czi", "versions", org, project, version)
```
jakeyheath, line 22:
```suggestion
	versionsPath := path.Join(os.UserHomeDirectory(), ".czi", "versions", org, project, version)
```

hvm/linkManager/set_bin_link.go
```diff
@@ -0,0 +1,72 @@
...
+	binPath := path.Join(home, ".czi", "bin")
+
+	os.MkdirAll(binPath, 0755)
+
+	if _, err := os.Stat(versionsPath); os.IsNotExist(err) {
+		fmt.Println("Requested version is not installed.")
+		return err
```
jakeyheath, line 29:
```suggestion
	if _, err := os.Stat(versionsPath); os.IsNotExist(err) {
		return errors.Wrap(err, "requested version is not installed")
```

jakeyheath: bubble up all the errors to one place and print them at the root of the application

shared/githubconnector/download.go
```diff
@@ -0,0 +1,62 @@
...
+	"path"
+)
+
+func (client *GithubConnector) DownloadPackage(org, project, version, os, arch, path string) (string, error) {
+
+	release, err := client.GetRelease(org, project, version)
+
+	if err != nil {
+		fmt.Println("Error getting release: ", err)
+		return "", err
```
jakeyheath, line 19:
```suggestion
		return "", errors.Wrap(err, "getting release")
```

shared/githubconnector/download.go
```diff
@@ -0,0 +1,62 @@
...
+	return "", errors.New("no suitable package found")
+}
+
+func download(url, fileName, dir string) (string, error) {
+
+	filePath := path.Join(dir, fileName)
+	// Download the url to the specified path/fileName
+	file, err := os.Create(filePath)
+	if err != nil {
+		return "", err
```
jakeyheath, line 37: can you wrap these errors so we know which statement errored?

shared/githubconnector/download.go
```diff
@@ -0,0 +1,62 @@
...
+	client := http.Client{
+		CheckRedirect: func(r *http.Request, via []*http.Request) error {
+			r.URL.Opaque = r.URL.Path
+			return nil
+		},
+	}
+
+	resp, err := client.Get(url)
+	if err != nil {
+		log.Fatal(err)
```
jakeyheath, line 49: let's bubble up this error, instead of using fatal. I think fatal causes the program to panic right here

shared/githubconnector/get_releases.go
```diff
@@ -0,0 +1,111 @@
...
+	}, nil
+
+}
+
+func (client *GithubConnector) GetReleases(org string, project string) ([]*Release, error) {
+
+	happyReleases := make([]*Release, 0)
+
+	// Only up to 1000 results are returned by the API
+	for page := 1; page < 2; page++ {
```
jakeyheath, line 35: do the pages work like this? I thought we had to parse the page out of the resp like

~~~
page = resp.NextPage
~~~

shared/githubconnector/get_releases.go
```diff
@@ -0,0 +1,111 @@
...
+	if len(parts) < 4 {
+		return ""
+	}
+
+	archAndExtension := strings.Split(name, "_")[3]
+	return strings.Split(archAndExtension, ".")[0]
+}
+
+func nameToOS(label string) string {
+	os := strings.Split(label, "_")[2]
```
jakeyheath, line 104:
```suggestion
	os := strings.SplitN(label, "_")[2]
```
I find split in to be safer since I usually know how many splits I am expecting. Also, we should make sure the split has at least 2 elements so we avoid a nil pointer error at runtime.

jakeyheath (approved):
love the tool! lots of little tiny nits, but overall great work

hvm/cmd/list-releases.go
dspadea: Woops. Yeah. The `use` command even went away. Will clean this up.

dspadea:
> love the tool! lots of little tiny nits, but overall great work

Thanks! I hope it's useful for everyone. And thanks for taking all the time for the details review!

shared/githubconnector/get_releases.go
dspadea: I'm only getting one page here. I was getting back all the responses, but they really weren't necessary, and I exhausted my API call limit for my IP the hour. The tool doesn't (yet?) authenticate with a PAT, so the limit is rather low.

Might later add an option to get more pages if people care.

As far as parsing the response, I'm using the Github client rather than a raw HTTP client, so I guess it's probably doing that for me. This is how the client I'm using works.

dspadea: Also, this will probably have to be configurable in the function call, since this is in the shared/ module and could be used for other stuff. Not going to bother with it just yet, though.

PR #2102, opened 2023-07-27
feat: Combine two different stack service implementations

alexlokshin-czi:
We currently have two divergent implementations of stack_service -- one in cli/pkg/stack_service, one in shared/stack; both were identical at some point, and both are being used in CLI through references, but have diverged quite extensively, each one maintaining their own mocks, tests, and tied into the code. This PR will combine both behaviors into a single implementation.

<details open>
  <summary><a href="https://czi-tech.atlassian.net/browse/CCIE-1636" title="CCIE-1636" target="_blank">CCIE-1636</a></summary>
  <br />
  <table>
    <tr>
      <th>Summary</th>
  <td>Combine two copies of stack_service into one (cli/pkg/stack_mgr and shared/stack)</td>
    </tr>
    <tr>
      <th>Type</th>
      <td>
        Task
        Task
      </td>
    </tr>
    <tr>
      <th>Status</th>
      <td>In Progress</td>
    </tr>
  </table>
</details>

---

Addresses [CCIE-1636]

[CCIE-1636]: https://czi-tech.atlassian.net/browse/CCIE-1636?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


shared/stack/stack_service.go
```diff
@@ -77,6 +76,12 @@ func (s *StackService) WithWorkspaceRepo(workspaceRepo workspacerepo.WorkspaceRe
 	return s
 }
 
+func (s *StackService) WithContext(env, appName string) *StackService {
```
jakeyheath, line 79: This function is misleading, because you don't pass in a context object. Why is this moved from the constructor? Both are required arguments for StackService.

alexlokshin-czi: One of the implementations has a WithHappyConfig(), which was removed and replaced with the new builder method. I can rename it to WithEnvContext. Passing these arguments into a constructor conflicts with the builder approach, which only see arguments get added over time.

jakeyheath: The builder approach still applies, but if the arguments are required, then we should keep them in the constructor. Otherwise you might build a bad object

alexlokshin-czi: Here's a list of args passed through builder calls:

WithBackend
WithExecutor
WithWorkspaceRepo

None of them are present in the constructor args. Creating a constructor with 5 args makes a messy constructor.

alexlokshin-czi: We can convert them to options.

alexlokshin-czi: But I'd prefer to do that in a separate PR.

shared/stack/stack.go
```diff
@@ -312,12 +331,22 @@ func (s *Stack) GetStackInfo(ctx context.Context) (*model.AppStackResponse, erro
...
+	}
+	// TODO: only here until people upgrade their CLIS. remove this in a few weeks
+	metaLegacy := StackMetaLegacy{}
+	err = json.Unmarshal(metaJSON, &metaLegacy)
+	if err != nil {
+		return nil, errors.Wrap(err, "could not unmarshal legacy stack meta")
+	}
+	err = meta.Merge(metaLegacy)
+	if err != nil {
+		return nil, errors.Wrap(err, "could not merge stack meta")
```
hspitzley-czi, line 349: I think it's been long enough that we don't need to port this over

shared/stack/stack_meta.go
```diff
@@ -34,6 +37,65 @@ type StackMeta struct {
...
+	Env       string `json:"happy/env"`
+	ImageTag  string `json:"happy/meta/imagetag"`
+	ImageTags string `json:"happy/meta/imagetags"`
+	App       string `json:"happy/app"`
+	Slice     string `json:"happy/meta/slice"`
+	CreatedAt string `json:"happy/meta/created-at"`
+	UpdatedAt string `json:"happy/meta/updated-at"`
+	Owner     string `json:"happy/meta/owner"`
+	Priority  string `json:"happy/meta/priority"`
+}
```
hspitzley-czi, line 97: Similarly, I think it's been long enough that we don't need to port this over

shared/stack/stack.go
```diff
@@ -20,6 +20,21 @@ import (
...
+	LastUpdated string            `json:"last_updated,omitempty"`
+	Message     string            `json:"message,omitempty"`
+	Outputs     map[string]string `json:"outputs,omitempty"`
+	Endpoints   map[string]string `json:"endpoints,omitempty"`
+	App         string            `json:"app,omitempty"`
+	GitRepo     string            `json:"git_repo,omitempty"`
+	GitSHA      string            `json:"git_sha,omitempty"`
+	GitBranch   string            `json:"git_branch,omitempty"`
+}
+
```
hspitzley-czi, line 37: This appears unused, can it be deleted?

hspitzley-czi (commented):
I think this is a dupe of https://github.com/chanzuckerberg/happy/pull/2096 with some additional builder expressions, I'm good with using either implementation but let's clean up some of the old stuff instead of porting it to the new implementation if we go with this PR

PR #2128, opened 2023-07-28
feat: side car env vars and HAPPY_CONTAINER

ademartini-czi:
<details open>
  <summary><a href="https://czi-tech.atlassian.net/browse/CCIE-1641" title="CCIE-1641" target="_blank">CCIE-1641</a></summary>
  <br />
  <table>
    <tr>
      <th>Summary</th>
  <td>HAPPY_CONTAINER env var</td>
    </tr>
    <tr>
      <th>Type</th>
      <td>
        Story
        Story
      </td>
    </tr>
    <tr>
      <th>Status</th>
      <td>To Do</td>
    </tr>
  </table>
</details>

---

Similar to https://github.com/chanzuckerberg/happy/pull/2084, but we also need to expose this information inside of sidecars so we can make use of it with tooling (ex: observability).

Also, HAPPY_CONTAINER will allow us to distinguish between different containers in the same pod, and filter accordingly in our Grafana dashboards.

https://czi-tech.atlassian.net/browse/CCIE-1641


jakeyheath (approved):

tonyta:
Suggesting that all env vars set by happy to be prefixed with `HAPPY_*` to avoid collision and clearly denote its origin.

PR #2190, opened 2023-08-02
feat: CCIE-1617: Initial commit of protos for Happy API and event system

dspadea:
<details open>
  <summary><a href="https://czi-tech.atlassian.net/browse/CCIE-1617" title="CCIE-1617" target="_blank">CCIE-1617</a></summary>
  <br />
  <table>
    <tr>
      <th>Summary</th>
  <td>Create protbuf (but not grpc) for Happy API Event schema</td>
    </tr>
    <tr>
      <th>Type</th>
      <td>
        Task
        Task
      </td>
    </tr>
    <tr>
      <th>Status</th>
      <td>Done</td>
    </tr>
  </table>
</details>

---

null


api/proto/hapi/generate.sh
```diff
@@ -0,0 +1,18 @@
...
+
+rm -rf $GO_DST_DIR $JS_DST_DIR $TS_DST_DIR
+mkdir  $GO_DST_DIR $JS_DST_DIR $TS_DST_DIR
+
+for proto in *.proto
+do
+    set -x
+    protoc -I=$SRC_DIR --go_out=$GO_DST_DIR $SRC_DIR/$proto
+    protoc -I=$SRC_DIR --grpc-web_out=import_style=typescript,mode=grpcwebtext:$TS_DST_DIR $proto
+    protoc -I=$SRC_DIR --js_out=import_style=commonjs:$JS_DST_DIR  --grpc-web_out=import_style=commonjs,mode=grpcwebtext:$JS_DST_DIR $proto
```
hspitzley-czi, line 16: I'm not sure we need JS in addition to TS?

api/proto/hapi/service-event.proto
```diff
@@ -0,0 +1,42 @@
...
+message HappyEvent {
+
+    /*   Happy Event ID
+
+        When creating a new notification, this should not be specified.
+        This field is populated when querying notification records from the
+        database.
+
+    */
+    optional string id = 1;
```
hspitzley-czi, line 33: Would `HappyEvent` be the struct that we store in the DB or are you thinking it would be wrapped in another struct that gets stored in the DB?

api/proto/hapi/config.proto
```diff
@@ -0,0 +1,45 @@
+syntax = "proto3";
+
+option go_package = "github.com/chanzuckerberg/happy/api/config";
+
+package happy.api.config;
```
hspitzley-czi, line 5: Can you think of a name besides `config` that would make sense here? I just want to avoid confusion with the `happy config` feature which we should probably make a proto for eventually

dspadea: Sure, will do.

api/proto/hapi/service-event.proto
dspadea: HappyEvent wraps Event and describes the record in the database. I'd initially combined them, but I got to thinking we can maybe reuse that core  Event to create something like OktaServiceEvent or whatever. That way it mates up, but we can add fields around it to publish on the bus.

When querying Happy API for event history, you'd get a list of HappyEvents back.

api/proto/hapi/generate.sh
dspadea: I wasn't sure which you needed. I can easily remove it if you don't want it. If we are going to publish it on npm's main repo, does it make sense to generate it? I'm not really a Node guy, so I'm not super familiar with how these things interoperate.

hspitzley-czi (approved):

api/proto/hapi/config.proto
dspadea: Renamed to `app`.

api/proto/hapi/service-event.proto
dspadea: I added `created_at` to the `HappyEvent` proto message. Seemed like it could be of interest to a potential calling application.

api/proto/hapi/generate.sh
dspadea: Per discussion, not sure what to do here. Going to kick this down the road to the packaging ticket. https://czi-tech.atlassian.net/browse/CCIE-1649

PR #2198, opened 2023-08-03
fix: send independent automated release announcements for happy cli and api

hspitzley-czi:
Make separate Jira versions and release announcements


jakeyheath (approved):
NICE

PR #2389, opened 2023-08-30
feat: CCIE-1649 package happy events objects in npm

dspadea:
<details open>
  <summary><a href="https://czi-tech.atlassian.net/browse/CCIE-1649" title="CCIE-1649" target="_blank">CCIE-1649</a></summary>
  <br />
  <table>
    <tr>
      <th>Summary</th>
  <td>Package Happy Events objects in NPM</td>
    </tr>
    <tr>
      <th>Type</th>
      <td>
        Task
        Task
      </td>
    </tr>
    <tr>
      <th>Status</th>
      <td>Done</td>
    </tr>
  </table>
</details>

---

Package and push to Github npm repo


.github/workflows/publish-hapi-api-protos.yml
```diff
@@ -0,0 +1,22 @@
...
+      - name: Install Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: '16.x'
+          registry-url: 'https://registry.npmjs.org'
+      - run: npm ci
+      - run: npm run build-proto
+      - run: npm version patch
+      - run: npm pack
+      - run: npm publish
```
hspitzley-czi, line 20: I think Travis said we should use github's npm registry for this, is that what this will do?

dspadea: Yeah, it should push to Github's npm. It's in the package.lock package configuration.

.github/workflows/publish-hapi-api-protos.yml
```diff
@@ -0,0 +1,22 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: '16.x'
+          registry-url: 'https://registry.npmjs.org'
+      - run: npm ci
+      - run: npm run build-proto
+      - run: npm version patch
```
hspitzley-czi, line 18: I think you'll want to have a step that parses the version from the new tag and use `npm version <newversion>` instead and you probably want to gate this so it only runs on releases of hapi-proto

hspitzley-czi: Here's an example of gating a job, this one only runs when the api is release https://github.com/chanzuckerberg/happy/blob/0b11e16b137506cb3e93f64f9802e898d1f7a8af/.github/workflows/api-release-prod.yaml#L11-L21

.github/workflows/publish-hapi-api-protos.yml
```diff
@@ -0,0 +1,46 @@
...
+    runs-on: [ARM64, self-hosted, Linux]
+    steps:
+      - name: Check Release
+        id: hapi_proto_released
+        uses: actions/github-script@v5
+        with:
+          result-encoding: string
+          script: return context.ref.includes('refs/tags/hapi-proto-v')
+    outputs:
+      hapi_proto_released: ${{ steps.api_released.outputs.result }}
```
hspitzley-czi, line 17:
```suggestion
      hapi_proto_released: ${{ steps.hapi_proto_released.outputs.result }}
```

.github/workflows/publish-hapi-api-protos.yml
dspadea: I've added gating logic and version parsing as above.

.github/workflows/publish-hapi-api-protos.yml
```diff
@@ -0,0 +1,46 @@
...
+  npm-publish:
+    runs-on:  [ARM64, self-hosted, Linux]
+    if: ${{ needs.check-protos-released.outputs.hapi_proto_released == 'true' }}
```
hspitzley-czi, line 20: I think you need a `needs` field here?
```suggestion
    if: ${{ needs.check-protos-released.outputs.hapi_proto_released == 'true' }}
    needs:
      - check-protos-released
```

release-please-config.json
```diff
@@ -81,7 +84,8 @@
         "api",
         "shared",
         "hvm",
-        "terraform-provider"
+        "terraform-provider",
+        "hapi-proto"
```
hspitzley-czi, line 88: Do these need to be linked?

If so that's ok but we'll need to be careful with the initial release to ensure it doesn't bump happy cli/api/...etc up to v1.0.0. We might wanna work through that initial release together.

dspadea: Ok, I'll remove it til later.

hspitzley-czi: I guess if happy api is going to use the protos they should probably be linked 🤔

release-please-config.json
```diff
@@ -26,6 +26,9 @@
     "api": {
       "package-name": "api"
     },
+    "hapi-proto": {
+      "package-name": "hapi-proto"
```
hspitzley-czi, line 30: Default is to release as `1.0.0`, let's pin this to `0.1.0` for now and then after the initial release we'll have to remove the pin and let it increment the minor version naturally
```suggestion
      "package-name": "hapi-proto",
      "release-as": "0.1.0"
```

hspitzley-czi (approved):

PR #2487, opened 2023-09-21
feat: cloudfront added to stack module

jakeyheath:
<details open>
  <summary><a href="https://czi-tech.atlassian.net/browse/CCIE-1853" title="CCIE-1853" target="_blank">CCIE-1853</a></summary>
  <br />
  <table>
    <tr>
      <th>Summary</th>
  <td>Cloudfront added to stack module</td>
    </tr>
    <tr>
      <th>Type</th>
      <td>
        Task
        Task
      </td>
    </tr>
    <tr>
      <th>Status</th>
      <td>In Progress</td>
    </tr>
  </table>
</details>

---

Ccie 1853

## Summary

This allows a user to configure vanity domains for their happy stacks. This is helpful for when a happy stack makes it to production and we want to have the user configure their domain name based on an external domain. By default, all stacks are deployed against the same Zone as part of the happy environment. This would allow an individual stack to say they want a different domain name. It is expected that the Route53 zone be configured out of band, such as in the long-lived infrastructure. Also, this process does not do the work of setting up the zone's nameservers. If nameservers need to be moved to a different vendor, like markmonitor, then we will have to do that manually.

## Caveats

* This really only works for service_type EXTERNAL. If INTERNAL is used, then the OIDC redirect will redirect back to the loadbalancer domain and not the cloudflare domain.
* Caching policies are limited at this time. Right now, we disable caching with the ability to override.
* This module should be attached to a specific stack out-of-band. Its too clunky to add to the stack's configuration, it should be added in the long-lived infra.


terraform/modules/happy-cloudfront/variables.tf
```diff
@@ -0,0 +1,67 @@
...
+    max_ttl     = optional(number, 300)
+    compress    = optional(bool, true)
+  })
+  description = "The cache settings for the CloudFront distribution."
+  default     = {}
+}
+
+variable "geo_restriction_locations" {
+  type        = set(string)
+  default     = ["US", "CA", "MX"]
```
abiju-czi, line 29: I don't think we need to add CA and MX for now

terraform/modules/happy-cloudfront/variables.tf
```diff
@@ -0,0 +1,67 @@
...
+  description = "The countries to whitelist for the CloudFront distribution."
+}
+
+variable "allowed_methods" {
+  type        = set(string)
+  default     = ["GET", "HEAD"]
```
abiju-czi, line 35: I think we might want to add POST and PUT?

jakeyheath: I should change this variable name. It is only related to cacheing allowed methods. All methods are allowed and forwarded

terraform/modules/happy-cloudfront/variables.tf
jakeyheath: good point

examples/typical_app/.happy/terraform/envs/rdev/main.tf
```diff
@@ -76,4 +84,9 @@ module "stack" {
   // tasks can be utilized to run post-deployment tasks such as database migrations or deletions
   tasks = {
   }
+
+
+  providers = {
```
jakeyheath, line 89: this is required for TLS certificates. They have to exist in this region for some reason

hspitzley-czi (approved):

PR #2510, opened 2023-09-26
feat: throw error to required key in get command

naihsuanshao:
This is for this [issue](https://czi-tech.atlassian.net/browse/CCIE-1846). This should throw error to notify user to provide the key when user run `happy config get`.

### DIFF
--- a/cli/cmd/config.go
+++ b/cli/cmd/config.go
@@ -17,6 +17,7 @@ import (
 
 var (
 	stack     string
+	getKey    string
 	fromEnv   string
 	fromStack string
 	logger    *logrus.Logger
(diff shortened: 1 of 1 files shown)


cli/cmd/config.go
```diff
@@ -35,6 +36,7 @@ func init() {
 	configListCmd.Flags().BoolVarP(&reveal, "reveal", "r", false, "Print the actual app config values instead of masking them")
 
 	configCmd.AddCommand(configGetCmd)
+	configGetCmd.Flags().StringVar(&getKey, "get-key", "", "Specify the key that you want to get")
```
jakeyheath, line 39: hmm this doesn't seem correct. I think the issue is in `var configGetCmd = &cobra.Command{` subcommand below (line149). I don't think its validating the correct number of args that are passed to `args`. When you fix this, can you add reproduction steps to your description?

jakeyheath (approved):
LGTM

Great description in the writeup, I really appreciated that.

cli/cmd/config.go
```diff
@@ -161,6 +161,13 @@ var configGetCmd = &cobra.Command{
 			return err
 		}
 
+		if len(args) == 0 {
+			noKeyProvidedMessage := messageWithStackSuffix(
+				fmt.Sprintf("Please supply the key name you want to look up."),
```
jakeyheath, line 166: I think you can use just a regular string here without the fmt.Sprintf

PR #2531, opened 2023-10-05
Make a datadog provider optional for happy-stack-eks

jgadling:
This is one way we can avoid requiring a datadog provider in the base module. Datadog is only required if dd dashboard or synthetics are enabled


jgadling:
Closing this since GHA is caught in an endless loop.

PR #2613, opened 2023-10-18
feat: fixed batch delete happy stack

naihsuanshao:
This is for slack request for batch deleting happy stack. See Jira issue [here](https://czi-tech.atlassian.net/browse/CCIE-1748).

Verified the change by `cd` into `examples/typical_app` folder then create two test stacks, then run `happy delete test-stack1 test-stack-2` to verify that the change works.

The for loop does iterate through all the arguments but one of the test stack deletion got terminated early when it ran into unauthorized error.

Used this launch.json in debugger to step through the code change
```
{
    // Use IntelliSense to learn about possible attributes.
    // Hover to view descriptions of existing attributes.
    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Launch Package",
            "type": "go",
            "request": "launch",
            "mode": "debug",
            "program": "../happy/cli",
            "args": ["delete", "test-nshao","-v"],
            "cwd": "../happy/examples/typical_app"
        }
    ]
}
```

### DIFF
--- a/cli/cmd/delete.go
+++ b/cli/cmd/delete.go
@@ -32,7 +32,16 @@ var deleteCmd = &cobra.Command{
 	Short:        "Delete an existing stack",
 	Long:         "Delete the stack with the given name.",
 	SilenceUsage: true,
-	RunE:         runDelete,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		fmt.Println("Running delete")
+		for _, stackName := range args {
+			err := runDelete(cmd, stackName)
+			if err != nil {
+				return err
+			}
+		}
+		return nil
+	},
 	PreRunE: happyCmd.Validate(
 		cobra.MinimumNArgs(1),
 		happyCmd.IsStackNameDNSCharset,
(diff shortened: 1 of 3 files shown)


cli/cmd/delete.go
```diff
@@ -32,7 +32,16 @@ var deleteCmd = &cobra.Command{
 	Short:        "Delete an existing stack",
 	Long:         "Delete the stack with the given name.",
 	SilenceUsage: true,
-	RunE:         runDelete,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		fmt.Println("Running delete")
```
alexlokshin-czi, line 36: Please change to our standard logging. Use of fmt.Print* is frowned upon.

cli/cmd/delete.go
```diff
@@ -32,7 +32,16 @@ var deleteCmd = &cobra.Command{
...
+		for _, stackName := range args {
+			err := runDelete(cmd, stackName)
+			if err != nil {
```
alexlokshin-czi, line 39: Might make sense to message and continue to the next stack.

cli/cmd/delete.go
naihsuanshao: Updated to using logrus.

cli/cmd/delete.go
naihsuanshao: Updated to log error and continue.

jakeyheath (approved):

PR #2636, opened 2023-10-24
fix: Docker compose validation misfires for containers that need to be built, but are not listed as services

alexlokshin-czi:
<details open>
  <summary><a href="https://czi-tech.atlassian.net/browse/CCIE-2030" title="CCIE-2030" target="_blank">CCIE-2030</a></summary>
  <br />
  <table>
    <tr>
      <th>Summary</th>
  <td>Happy: Docker compose validation misfires for containers that need to be built, but are not listed as services</td>
    </tr>
    <tr>
      <th>Type</th>
      <td>
        Bug
        Bug
      </td>
    </tr>
    <tr>
      <th>Status</th>
      <td>In Review</td>
    </tr>
  </table>
</details>

---

null


hspitzley-czi (approved):

cli/cmd/init.go
```diff
@@ -301,49 +301,50 @@ func validateConfigurationIntegirty(ctx context.Context, slice string, happyClie
...
 		ss := sets.NewStringSet().Add(configServices...)
-		for serviceName, service := range availableServices {
-			// ignore services that don't have a build section
-			// as these are not deployable services
-			if service.Build == nil {
-				continue
-			}
-			ok := ss.ContainsElement(serviceName)
+		for _, serviceName := range configServices {
+			service, ok := composeServices[serviceName]
```
jakeyheath, line 324: thinking...

I had this loop like this for a different bug but trying to remember what it was

jakeyheath: ooo slices. that is why. This approach doesn't work when you are working with slices. See my comment:
```

// NOTE: availableServices will only contain the services that are a part of the slice.
		// That means we have to iterate over those first to make sure they are all in the config.json and
		// not the other way around.
```

jakeyheath: To test this out, I would borrow one of the single-cell apps and make sure your validation works as expected when using their slices

alexlokshin-czi: @jgadling issue is the exact opposite issue -- for a non-slice setup a service that has a build section in docker-compose but is not a happy service, validation fails.

jgadling: I don't know what this means, we're not using slices, but I'm inferring from your comment that I can work around the issue I'm having by removing the (unnecessary) "build" section from the second service in my docker-compose file.

jakeyheath (approved):

PR #2644, opened 2023-10-26
fix: Update the pin to avoid caching

alexlokshin-czi:


examples/integration_test/.happy/terraform/envs/rdev/main.tf
```diff
@@ -1,5 +1,5 @@
 module "stack" {
-  source = "git@github.com:chanzuckerberg/happy//terraform/modules/happy-stack-eks?ref=main"
+  source = "git@github.com:chanzuckerberg/happy//terraform/modules/happy-stack-eks?ref=v0.114.0"
```
jakeyheath, line 2: don't we always want latest so we can test whatever is newest?

PR #2776, opened 2023-11-20
feat: Happy module parser should be able to read from both local reference or remote happy stack

naihsuanshao:
<details open>
  <summary><a href="https://czi-tech.atlassian.net/browse/CCIE-2102" title="CCIE-2102" target="_blank">CCIE-2102</a></summary>
  <br />
  <table>
    <tr>
      <th>Summary</th>
  <td>ERROR message about HCL validation</td>
    </tr>
    <tr>
      <th>Type</th>
      <td>
        Task
        Task
      </td>
    </tr>
    <tr>
      <th>Status</th>
      <td>In Review</td>
    </tr>
  </table>
</details>

---

This change is for a confusing error log from Terrform validate step for Happy Cli. Currently, the validator would skip over local reference of module, and log this error at this [line](https://github.com/chanzuckerberg/happy/blob/main/shared/hclmanager/hcl_manager.go#L164).

I moved things around and concatenated the path under the assumption that all happy project should follow the same directory system.

PR ref: https://czi-tech.atlassian.net/browse/CCIE-2102

I verified the change by running the `happy list` against `typical_app` in Happy project and `single-cell-data-portal`, where `typical_app` is using the remote reference and `single-cell-data-portal` is using local reference.

I've attached my `launch.json`.
```
{
    // Use IntelliSense to learn about possible attributes.
    // Hover to view descriptions of existing attributes.
    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Launch Package",
            "type": "go",
            "request": "launch",
            "mode": "debug",
            "program": "../happy/cli",
            "args": ["list", "-v"],
            // "cwd": "../happy/examples/typical_app"
            "cwd": "../single-cell-data-portal"
        }
    ]
}
```


shared/util/tf/tf_parser.go
```diff
@@ -190,24 +190,25 @@ func (tf TfParser) ParseModuleCall(happyProjectRoot, tfDirPath string) (ModuleCa
...
-			if err != nil {
-				return fmt.Errorf("%w: %w", err, ErrUnableToDownloadModuleSource)
+				// Download the module source
+				err = getter.GetAny(sourceDir, source.AsString())
+				if err != nil {
+					return fmt.Errorf("%w: %w", err, ErrUnableToDownloadModuleSource)
+				}
+			} else {
+				// local reference
+				sourceDir = fmt.Sprintf("%s%s%s", happyProjectRoot, "/.happy/terraform/", source.AsString()[2:])
```
alexlokshin-czi, line 208: This works for source set to `git@github.com:chanzuckerberg/happy//terraform/modules/happy-stack-eks?ref=main` and `./modules/ecs-stack`, but what about `./modules/happy-stack-eks`? Is it possible to determine whether the source is remote or local by looking at the source structure? Something like - if it starts with a period, it's local, if it starts with `git@` or `https:` or contains `github.com:` it's remote?

naihsuanshao: Good call. I've changed the logic of checking local reference. I think all our remote reference starts with `git@`, but I could also check for `https:` if that would be more safe.

alexlokshin-czi: I don't mean to drag it out :) Let's create a method that determines if a path is local, and use if for both tf_parser and hcl_manager. Also, let's check on to determine a remote source 

```
			if strings.HasPrefix(source.AsString(), "git@") || strings.HasPrefix(source.AsString(), "https:") {
```

naihsuanshao: No worries! Let me added this and verify that it works!

naihsuanshao: Move the logic into a shared function reside in `tf_generatator.go`, and added `http`

alexlokshin-czi: I came across the terraform doc on this subject, it's more complex than I thought :) https://developer.hashicorp.com/terraform/language/modules/sources . Based on what we're dealing with, maybe, it's easier to determine the locally sourced modules. Per doc, it is 

```
A local path must begin with either ./ or ../ to indicate that a local path is intended, to distinguish from a module registry address.
```

jakeyheath:
stale

PR #2928, opened 2024-01-08
fix: Fix integration test (it requires eks:listclusters)

alexlokshin-czi:


.github/workflows/integration-test.yml
```diff
@@ -2,6 +2,7 @@ on:
   push:
     branches:
       - main
+      - alexlokshin-czi-patch-2
```
hspitzley-czi, line 5:
```suggestion
```

hspitzley-czi (approved):

PR #2937, opened 2024-01-09
fix: only get stack info on delete if deleting images

jakeyheath:
<details open>
  <summary><a href="https://czi-tech.atlassian.net/browse/CCIE-2231" title="CCIE-2231" target="_blank">CCIE-2231</a></summary>
  <br />
  <table>
    <tr>
      <th>Summary</th>
  <td>Delete is bailing too quickly when state isn't there</td>
    </tr>
    <tr>
      <th>Type</th>
      <td>
        Task
        Task
      </td>
    </tr>
    <tr>
      <th>Status</th>
      <td>In Progress</td>
    </tr>
  </table>
</details>

---

## Summary

This fixes a bug where the stacks couldn't be deleted if there was no statefile yet for the stack. This was due to a feature flag added. This fix moves the logic for the feature flag so that delete's no longer accidently bail when deleting stacks without statefiles.


cli/go.mod
```diff
@@ -1,4 +1,6 @@
-go 1.20
+go 1.21
```
hspitzley-czi, line 1: Is this upgrade intended as part of this?

hspitzley-czi: I would advocate for taking this on later and doing all happy packages at the same time

jakeyheath: not sure how that happened, I will revert

hspitzley-czi: `gvm` (if you use that) has a tendency to lose track of which go version it's supposed to be on :/

hspitzley-czi (approved):

PR #3015, opened 2024-02-08
fix: Happy doesn't cache module code, and keeps pulling identical modules on every run during HCL verification

alexlokshin-czi:


jakeyheath (approved):
Can you post a speedup time in this PR description?

PR #3586, opened 2024-10-02
feat: allow k6 operator service account in rdev/staging (CCIE-3437)

christophermanahan:
<details open>
  <summary><a href="https://czi.atlassian.net/browse/CCIE-3437" title="CCIE-3437" target="_blank">CCIE-3437</a></summary>
  <br />
  <table>
    <tr>
      <th>Summary</th>
  <td>happy edu - Deploy and test k6 in staging</td>
    </tr>
    <tr>
      <th>Type</th>
      <td>
        Story
        Story
      </td>
    </tr>
    <tr>
      <th>Status</th>
      <td>To Do</td>
    </tr>
  </table>
</details>

---

Ticket: https://czi.atlassian.net/browse/CP-1504

Doc: https://czi.atlassian.net/wiki/spaces/ETE/pages/3192881220/PEMA+Project+End+Manual+Approvals

While [Grafana k6](https://grafana.com/docs/k6/latest/) tests have been successfully POC’d in rdev via injection of the LinkerD sidecar into the test runner pods that are deployed by the k6 operator to test our private backend service endpoints - staging introduces LinkerD mesh access control lists that prevent the test runner pod from making successful requests to our services in staging. This PR adds the service account `k6-operator-controller` under the namespace `k6-operator-system` to the staging and rdev environments so that the test runner pods deployed by the k6 operator under the `k6-operator-controller` service account can make successful requests to the services in rdev and staging. 

A test stack of the session service was deployed to staging with the happy config pointing to the git SHA associated with this PR and a successful test run to create a session was achieved in staging to validate this approach and change.

Screenshot 2024-10-02 at 12 27 52 PM


terraform/modules/happy-service-eks/variables.tf
```diff
@@ -296,6 +296,7 @@ variable "routing" {
     success_codes : optional(string, "200-499")
     service_type : string
     service_mesh : bool
+    allow_k6_operator : bool
```
jakeyheath, line 299:
```suggestion
    allow_k6_operator : optional(bool, true)
```

jakeyheath: make match the default above

christophermanahan:
> @christophermanahan instead of checking if you're in rdev or staging, do you think you can add another variable that's like, allow_k6_operator_controller with a description about grafana k6s? Then set the "default" to false so you can turn it on using that variable.
> 
> That way, we don't have the brittle logic about whether it's rdev or staging. Just something like this:
> 
> ```
> k6_operator_service_account = var.allow_k6_operator_controller ? [{
>     "kind"      = "ServiceAccount"
>     "name"      = "k6-operator-controller"
>     "namespace" = "k6-operator-system"
>   }] : []
> ```
> 
> more about variables here: https://developer.hashicorp.com/terraform/language/values/variables We have variables.tf for blocks like this

Updated to be a variable, thank you for the review and suggestion!

